PR #4, opened 2021-05-17
Sched 25

stroncod:
Add Resource service mock to GreedyMax.


sraaphorst:
I don't think we want pycache files in the repo, so we should probably add pycache to the `.gitignore` file.

resource_mock/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+        out_dict = {}
+        with open(f'{self.path}/gmos{site}_fpu_barcode.txt') as f:
+            
+            for row in f:
+                fpu, barcode = row.split()
+                out_dict[fpu] = barcode
+        return out_dict
+    
+    def _to_bool(self, b: str) -> bool:
+        return True if b == 'Yes' else False
```
sraaphorst, line 48: Again, this can just be `return b == 'Yes'`

resource_mock/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+    def _excel_reader(self, sites: List[str]) -> NoReturn:
+        workbook = load_workbook(filename=f'{self.path}/2018B-2019A Telescope Schedules.xlsx')
+        for site in sites:
+            sheet = workbook[site]
+            self.instruments[site.lower()] = {row[0].value.strftime("%Y-%m-%d"): 
+                                              [c.value for c in row[3:] ] for row in sheet.iter_rows(min_row=2)}
+
+            self.mode[site.lower()] = {row[0].value.strftime("%Y-%m-%d"): row[1].value for row in sheet.iter_rows(min_row=2)}
+            self.LGS[site.lower()] = {row[0].value.strftime("%Y-%m-%d"): 
+                                      self._to_bool(row[2].value) for row in sheet.iter_rows(min_row=2)}
```
sraaphorst, line 59: You use `site.lower()` here multiple times, requiring three calls to calculate the same thing.
Change this to:
```
site_lower = site.lower()
self.mode[site_lower] = ...
...
```
so we only calculate it once and get rid of the repetition.

Same with `row[0].value.strftime("%Y-%m-%d")`:
```
date = row[0].value.strftime('%Y-%m-%d')
```
and then use `date` for the `dict` keys.

resource_mock/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+            
+            for row in f:
+                fpu, barcode = row.split()
+                out_dict[fpu] = barcode
+        return out_dict
+    
+    def _to_bool(self, b: str) -> bool:
+        return True if b == 'Yes' else False
+
+    def _excel_reader(self, sites: List[str]) -> NoReturn:
```
sraaphorst, line 50: It would be better if we did a:
```
from greedy_max import Site
...
def _excel_reader(self, sites: List[Site]) -> NoReturn:
```

resource_mock/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+        workbook = load_workbook(filename=f'{self.path}/2018B-2019A Telescope Schedules.xlsx')
+        for site in sites:
+            sheet = workbook[site]
```
sraaphorst, line 53: Then we can get the name of the site as a string using `site.name.lower()`.

resource_mock/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+            raise Exception("Problems on reading spreadsheet...") 
+
+    def connect(self, sites: List[str]) -> NoReturn:
+        """
+        Allows the mock to load all the data locally, emulating a connection to the API.
+
+        """
+        for site in sites:
+
+            _site = 'gs' if site == 's' else 'gn'
```
sraaphorst, line 71: `_site = Site.GS if site == 's' else Site.GN`

resource_mock/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+            self.IFU[_site]['FPU'], self.FPU[_site] = self._load_fpu('FPU',site)
+            self.IFU[_site]['FPUr'], self.FPUr[_site] = self._load_fpu('FPUr',site)
+            self.GRAT[_site] = self._load_grat(site)
+            self.fpu_to_barcode[_site] = self._load_f2b(site)
+
+        if not self.FPU or not self.FPUr or not self.GRAT:
+            raise Exception("Problems on reading files...") 
+        
+        self._excel_reader(['GS','GN'])
```
sraaphorst, line 80: Do we need to pass these to `_excel_reader` if they are always the same?

sraaphorst: I would change `_excel_reader` to just take `self` and then as the first line, do:
```
sites = [site.name for site in Site]
```
and you will get the same list of strings, but like I said, we should try to use `Enum` or `IntEnum` where possible when there is a set, limited number of possibilities (like `Site` and `Band`).

resource_mock/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def night_info(self, info: str, site: str, date: str):
```
sraaphorst, line 82: Again, `site: Site`?

resource_mock/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+        elif info == 'LGS':
+            return self.LGS[site][date] if date in self.LGS[site] else None
+        
+        elif info == 'mode':
+            return self.mode[site][date] if date in self.mode[site] else None
+        
+        elif info == 'fpu-ifu':
+            return self.IFU[site]['FPU'][date] if date in self.IFU[site]['FPU'] else None
+        elif info == 'fpur-ifu':
+            return self.IFU[site]['FPUr'][date] if date in self.IFU[site]['FPUr'] else None
```
sraaphorst, line 105: This could be simplified by doing:
```
info_types = { 'fpu': self.fpu[site], fpur: self.fpur[site], 'instr': self.instruments[site], 'LGS': self.lgs[site], 'mode': self.mode[site], 'fpu-ifu': self.ifu[site]['FPU'], 'fpur-ifu': self.ifu[site]['FPUr'] }
```

and then:
```
if info in info_types:
     return info_types[info][date] if date in info_types[info]
else:
     print(f"No information about {info} is stored')
     return None
```

solver_greedy_max.py
```diff
@@ -8,6 +8,9 @@
 from astropy.visualization import time_support
 time_support()
 
+from resource_mock import Resource
```
sraaphorst, line 11: Any imports should always be before function calls, so you should swap this with `time_support()`, as this is Python standard.

solver_greedy_max.py
```diff
@@ -19,11 +22,33 @@
...
+    
+    fpu = {site: resource.night_info('fpu', site.name.lower(), night_date) for site in sites}
+    fpur = {site: resource.night_info('fpur', site.name.lower(), night_date) for site in sites}
+    grat = {site: resource.night_info('grat', site.name.lower(), night_date) for site in sites}
+    instruments = {site: resource.night_info('instr', site.name.lower(), night_date) for site in sites}
+    lgs = {site: resource.night_info('LGS', site.name.lower(), night_date) for site in sites}
+    modes = {site: resource.night_info('mode', site.name.lower(), night_date) for site in sites}
+    ifus = {'FPU':None,'FPUr':None}
+    ifus['FPU'] = {site: resource.night_info('fpu-ifu', site.name.lower(), night_date) for site in sites}
+    ifus['FPUr'] = {site: resource.night_info('fpur-ifu', site.name.lower(), night_date) for site in sites}
```
sraaphorst, line 43: Again, better to calculate `site.name.lower()` only once at the top:
`site_lower = site.name.lower()`
and then replace these.

The pattern:
```
variable = {site: resource.night_info(name, site_lower, night_date) for site in sites}
```
can definitely be simplified down to a couple lines. I would change this to (dunno if the dict type is str):
```
def _get_info(info_name: str, site: List[Sites], night_date: str) -> Dict[Site, str]:
     return {site: resource.night_info(info_name, site.name.lower(), night_date) for site in sites }
```

and then replace all of the calls with, for example:
```
fpu = _get_info('fpu', sites, night_date)
```
to avoid the repetition and make it more readable.

sraaphorst, line 43: This can definitely be simplified:

```
def _get_info(info_name, sites: List[Site], night_date) -> Dict[Site, str]:
     return { site: resource.night_info(info_name, site.name.lower, night_date) for site in sites }
```

and then replace each of the above with the equivalent of this:
```
fpu = _get_info('fpu', sites, night_date)
```

solver_greedy_max.py
```diff
@@ -43,7 +68,6 @@
    
     sites_by_obs = {obs_id: 'gs' if sum(ttab_gngs['weight_gs'][idx]) > 0 else 'gn' for idx, obs_id in enumerate(obs_ids)}
```
sraaphorst, line 69: I'd still like to see this changed to `Site.GS` and `Site.GN` instead of the strings.

sraaphorst (changes requested):
Good, but some improvements / simplifications could be made.

Also, we should add:
```
# Byte-compiled / optimized / DLL files
__pycache__/
*.py[cod]
```
to `.gitignore` to get rid of the `__pycache__` files, which shouldn't be in github.

greedy_max/schedule.py
```diff
@@ -51,10 +51,19 @@ def calibrate(self, dt):
...
+    def __init__(self, time_slot_length, weights, total_amount, 
+                 fpu, fpur, grat, instruments, lgs, mode, fpu2b, ifus):
         self.slot_length = time_slot_length
         self.weights = weights
         self.total = total_amount
+        self.fpu = fpu
+        self.fpur = fpur
+        self.grating = grat 
+        self.instruments = instruments
+        self.LGS = lgs
```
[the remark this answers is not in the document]
stroncod: I'm sure of the full acronym:  "laser guide ..." ?

greedy_max/schedule.py
```diff
@@ -70,4 +79,34 @@ def intervals(self, empty_slots: np.ndarray) -> np.ndarray:
...
-        return idx
+        return idx 
+
+    def _decode_mask(self, mask_name):
+        decoder = {'A':'0','B':'1','Q':'0',
+                                 'C':'1','LP':'2','FT':'3',
+                                 'SV':'8','DD':'9'}
+        pattern = '|'.join(map(re.escape, decoder.keys()))
+        return '1'+ re.sub(f'({pattern})', 
+                           lambda m: decoder[m.group()], mask_name).replace('-','')[6:]
```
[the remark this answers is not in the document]
stroncod: Transform the mask name (a string) to a barcode (another string): 
GN2003BQ034-01 -> 11003401

resource_mock/__init__.py
stroncod: I removed the comprehension for the dictionaries to apply this one

solver_greedy_max.py
stroncod: It's a great idea. The return type might be a bit complicated because it depends on what information is retrieved. Do you know how to handle this?

greedy_max/schedule.py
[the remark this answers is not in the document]
stroncod: Found it lol Laser guide star

greedy_max/schedule.py
```diff
@@ -70,4 +79,34 @@ def intervals(self, empty_slots: np.ndarray) -> np.ndarray:
...
         idx = np.digitize(cvec, bins=np.arange(ni) + 1)
 
-        return idx
+        return idx 
+
+    def _decode_mask(self, mask_name):
+        decoder = {'A':'0','B':'1','Q':'0',
+                                 'C':'1','LP':'2','FT':'3',
+                                 'SV':'8','DD':'9'}
+        pattern = '|'.join(map(re.escape, decoder.keys()))
```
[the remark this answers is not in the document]
stroncod: Actually it work! I guess there is not need to escape if the input is controlled and does not need that.

greedy_max/schedule.py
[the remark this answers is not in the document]
sraaphorst: Haha I should probably have mentioned that :-)

solver_greedy_max.py
sraaphorst: Sure. If you have possible multiple return types, you can use `Either`, like `Either[str, int, Observation]`.

PR #7, opened 2021-07-01
Sched 28: Declutter of __init__.py and other fixes

stroncod:
To reduce the size of __init__.py the following changes were applied:

- Both Plan and TimeWindow classes were removed from the main file and moved to a file of their own.
- Utils.py file was created to contain all static methods the greedy_max module uses
- Some logic from `_insert` method was moved into two new methods: `_integrate` and `_calibrate` to expand modularity in the main file 
-  Minor fixes on `_insert` were discover and fixed on not tested branches in the code. 
- `_interval` method was replace so it would be easier to understand and maintain.


sraaphorst (approved):
This all looks fine to me!

PR #10, opened 2021-07-14
Sched 30-29: Fix missing information in Resource and change naming convention to Visit

stroncod:
This PR contains both SCHED 30 and 29.

- On the Resource Mock:
   - There was an error at reading the excel file that caused missing information
   -  Information of missing dates is retrieve from the nearest date before the asking one 
- On the Naming convention:
   - The `Scheduling Unit` class was renamed `Visit` in according with the correct definition.

### DIFF
--- a/greedy_max/__init__.py
+++ b/greedy_max/__init__.py
@@ -13,10 +13,10 @@ stream_handler.setLevel(logging.INFO)
 logger.addHandler(stream_handler)
 
 class GreedyMax:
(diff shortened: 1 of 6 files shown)


sraaphorst (commented):
We should consider extracting this to a static internal method or something similar since we use it three times in a row, I think.

sraaphorst (approved):
Looks good to me.

stroncod:
> We should consider extracting this to a static internal method or something similar since we use it three times in a row, I think.

Where exactly ? I don't see what line you are referring to

PR #16, opened 2021-09-22
SCHED-44: Cleanup and bugfixes

sraaphorst:
A variety of cleanup and bugfixes to the Scheduler.
I am trying to make the code more modular, portable, and clean, and have tested before each commit to make sure that we get the same results and I haven't broken anything.

I also cleaned up some variable names for clarity and tried to get rid of as many warnings as PyCharm's use of pylint was flagging (e.g. variable names in Python and function names should not be capitalized or camel case, such as `GetProgramId` should be `get_program_id` and `programId` should be `program_id`). These are recommended Python standards according to the Python website.


collector/__init__.py
```diff
@@ -447,18 +426,18 @@ def _process_observation_data(self,
...
+                collected_program.add_observation(self.observation_num)
 
 
                 # Check if group exists, add if not
                 if group['key'] not in self.scheduling_groups.keys():
                     self.scheduling_groups[group['key']] = {'name': group['name'], 'idx': []}
-                    collected_program.add(group['key'])
-                self.scheduling_groups[group['key']]['idx'].append(self.nobs)
+                    collected_program.add_group(group['key'])
+                self.scheduling_groups[group['key']]['idx'].append(Collector.observation_num)
```
stroncod, line 439: Why here is `Collector.` but above in 433 is `self.` being used?

collector/__init__.py
```diff
@@ -17,89 +17,91 @@
...
+    # Counter to keep track of observations.
+    observation_num = 0
+
+    def __init__(self,
+                 sites: FrozenSet[Site],
+                 semesters: FrozenSet[str],
+                 program_types: FrozenSet[str],
+                 obs_classes: FrozenSet[str],
+                 time_range: Time = None,
+                 delta_time: Time = 1.0 * u.min):
```
stroncod, line 86: I know Bryan uses `delta_time` but i think `time_slot_length` is what we use in Greedymax and I think is more descriptive of what is really is.

common/structures/site.py
```diff
@@ -0,0 +1,24 @@
...
+
+# Dict from Site to geographical location.
+GEOGRAPHICAL_LOCATIONS = {Site.GS: EarthLocation.of_site('gemini_south'),
+                          Site.GN: EarthLocation.of_site('gemini_north')}
+
+# Site abbreviations for FPUs, gratings, etc.
+SITE_ABBREVIATION = {Site.GS: 'S', Site.GN: 'N'}
+
+# Site zip timestamps.
+SITE_ZIPS = {Site.GS: "-0830.zip", Site.GN: "-0715.zip"}
```
stroncod, line 24: Maybe Timezones should go here too?

resource_mock/__init__.py
```diff
@@ -77,25 +76,25 @@ def _excel_reader(self) -> NoReturn:
                 date = row[0].value
                 self.instruments[site][date] = [c.value for c in row[3:]] 
                 self.mode[site][date] = row[1].value
-                self.lgs[site][date] = self._to_bool(row[2].value)
+
+                # TODO: Some of these rows have None as their value. Is this right?
```
stroncod, line 82: I'll take a look at that

selector/__init__.py
```diff
@@ -470,7 +468,7 @@ def select(self, visits: List[Visit],
                         wind_conditions = actual_wind_conditions.get_wind_conditions(obs.visibility.azimuth[inight])
                     
                     # NOTE: first condition always going to be true. Why is needed?
-                    if too_status != 'rapid' and obs.too_status != 'none':
+                    if too_status != ToOType.RAPID and obs.too_status != ToOType.NONE:
```
stroncod, line 471: There is a need for `ToOType.None` instead of just using `None`?

stroncod (approved):
Add a couple of questions but it looks good. I'll get it ready for the merge

collector/__init__.py
```diff
@@ -110,7 +112,7 @@ def __init__(self, sites: List[Site],
         self.obsid = []
         self.obstatus = []
         self.obsclass = []
-        self.nobs = 0
+        self.observation_num = 0
         self.band = []
         self.toostatus = []
```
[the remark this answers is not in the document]
sraaphorst: Thanks for letting me know. I guess we'll use these in the future?
I'm just checking to make sure they're initialized properly and named logically and things flow cohesively.

collector/__init__.py
```diff
@@ -231,18 +233,18 @@ def _instrument_setup(self, configuration: Dict[str, List[str]], instrument_name
...
         
     def _readzip(self, 
                  zipfile: str, 
                  xmlselect: List[str], 
-                 site: str,
-                 selection=['ONGOING', 'READY'], 
-                 obsclasses=['SCIENCE'], 
+                 site: Site,
+                 selection=frozenset(['ONGOING', 'READY']),
+                 obsclasses=frozenset(['SCIENCE']),
```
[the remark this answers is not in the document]
sraaphorst: A list would do the trick, but a list is mutable, and  has a longer search time than a frozenset. pylint was complaining about passing mutable lists to functions, and since I know these both can be treated as sets and should be immutable, I decided to make them frozensets to eliminate the warning and make the API less fragile.

collector/__init__.py
sraaphorst: Ahhh... so my understanding is that `nobs` was a placeholder for `number_of_observations`.
We can make this a static member instead of a `Collector` implementation specific number, in which case, we **can** access it through `self`, but that's simply deceptive because `observation_num` now belongs to all `Collector` instances.

Do you foresee any problems with that? I couldn't, but I still have lots of reading to do. :-)

collector/__init__.py
sraaphorst: Sure. I'll change that!

common/structures/site.py
sraaphorst: I think this is a good idea. I'll do that as well.

selector/__init__.py
sraaphorst: I think it's just a matter of semantics. If `too_status` is `None`, that means that it could not be read.
If it is `ToOType.NONE`, that means that it was read correctly and was deliberately specified as `ToOType.NONE`, so there is a subtle distinction.

sraaphorst:
I'll merge now and create another PR to cover your ideas.

PR #17, opened 2021-09-23
SCHED-44: Continuation of cleanup

sraaphorst:
A lot of this is me going through PyCharm and pylint, cleaning up warnings and looking for improvements or bugfixes on the way, so some of the reformatting can be ignored: it is just PyCharm conforming to Python standards as set in the docs.

There are some bugfixes, code enhancements, etc.


collector/__init__.py
```diff
@@ -472,40 +473,38 @@ def create_time_array(self):
...
-                                       selected_variant['wv'])
-
-        # TODO: time_blocks expects float, gets List[Time] instead.
-        actcond['wind'] = WindConditions( selected_variant['wsep'], selected_variant['wspd'],
-                             selected_variant['wdir'], time_blocks)
-        
-        return actcond
+        sky = SkyConditions(SB.SBANY, selected_variant['cc'], selected_variant['iq'], selected_variant['wv'])
+        wind = WindConditions(selected_variant['wsep'], selected_variant['wspd'], selected_variant['wdir'], time_blocks)
+        return Conditions(sky, wind)
```
stroncod, line 510: Nice!

selector/__init__.py
```diff
@@ -313,215 +324,207 @@ def visibility(self, site: Site, jobs=0, ephem_path=None, overwrite=False, sbtwo
...
-              
+            for obs in tqdm(observations):
                 res.append(self._calculate_visibility(site,
-                                                      observations[id].target.designation,
-                                                      observations[id].target.tag,
-                                                      observations[id].target.coordinates,
-                                                      observations[id].sky_conditions,
-                                                      observations[id].elevation,
-                                                      obs_windows[id],
+                                                      observations[obs.idx].target.designation,
```
stroncod, line 329: Better just use `obs.target.designation`

stroncod (approved):
Minor comments on the collector and selector.

selector/__init__.py
sraaphorst: Good point. With the loop change, for all of these we can just use `obs` instead of `observations[obs.idx]`.

collector/__init__.py
```diff
@@ -130,47 +131,47 @@ def load(self, path: str) -> NoReturn:
         # Retrieve the site details
         try:
             site = Site(site_name)
-            location = GEOGRAPHICAL_LOCATIONS[site]
+            self.locations[site] = GEOGRAPHICAL_LOCATIONS[site]
```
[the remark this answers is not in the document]
sraaphorst: What will happen, though, when we allow for multisite? Since the Selector uses these values in the Collector, I am going to make a note and leave this one as is for now, I think... that way we can still have the `try`.

PR #20, opened 2021-09-25
Sched 43: Modify new GreedyMax algorithms to match original algorithm results

stroncod:
Fixes a bunch of bugs that caused missing Visits in the selection process

### DIFF
--- a/collector/__init__.py
+++ b/collector/__init__.py
@@ -375,7 +375,7 @@ class Collector:
                 if not conditions or conditions is None:
                     sky_cond = SkyConditions()
                 else:
-                    parse_conditions = conditions_parser(conditions)
+                    parse_conditions = conditions_parser(conditions)                    
                     sky_cond = SkyConditions(*parse_conditions)
 
                 # Elevation constraints        
--- a/collector/__init__.py
+++ b/collector/__init__.py
@@ -347,12 +347,13 @@ class Collector:
                 target_designation = None
                 target_tag = None
                 if targets:
-                    for target in targets:
+                    for target in targets:                        
                         try:
                             target_name = target['group']['name']
-                            target_tag = TargetTag(target['tag']) if target_name == 'Base' else None
-                            target_designation = target['num'] if target_tag is not None and \
-                                                                  target_tag == TargetTag.MajorBody else target['des']
+                            if target_name == 'Base':
+                                target_tag = TargetTag(target['tag'])
+                                if target_tag is not TargetTag.Sidereal and target_tag is not None:
+                                    target_designation = target['num'] if target_tag is TargetTag.MajorBody else target['des']
                         except:
                             pass
 
--- a/collector/__init__.py
+++ b/collector/__init__.py
@@ -371,7 +371,7 @@ class Collector:
                 # Sky Conditions
                 conditions = get_conditions(raw_observation, label=False)
 
-                if conditions or conditions is None:
+                if not conditions or conditions is None:
                     sky_cond = SkyConditions()
                 else:
                     parse_conditions = conditions_parser(conditions)
--- a/greedy_max/schedule.py
+++ b/greedy_max/schedule.py
@@ -164,11 +164,11 @@ class Visit:
         Use the most restrictive value for each condition. 
         '''
         restrictive_iq = min([obs.sky_conditions.iq for obs in self.observations])
-        restrictive_bg = min([obs.sky_conditions.sb for obs in self.observations])
+        restrictive_sb = min([obs.sky_conditions.sb for obs in self.observations])
         restrictive_cc = min([obs.sky_conditions.cc for obs in self.observations])
         restrictive_wv = min([obs.sky_conditions.wv for obs in self.observations])
 
-        return SkyConditions(restrictive_iq,restrictive_bg,restrictive_cc,restrictive_wv)
+        return SkyConditions(restrictive_sb,restrictive_cc,restrictive_iq,restrictive_wv)
             
 
     def __contains__(self, obs_idx:int) -> bool:
--- a/resource_mock/__init__.py
+++ b/resource_mock/__init__.py
@@ -50,11 +50,6 @@ class Resource:
                 out_dict[fpu] = barcode
         return out_dict
 
-    # TODO: Does not appear to be in use, and can be static.
-    @staticmethod
-    def _nearest(items, pivot):
-        return min(items, key=lambda x: abs(x - pivot))
-
     @staticmethod
     def _previous(items, pivot):
         # Return date equal or previous to pivot
@@ -118,12 +113,7 @@ class Resource:
         if info in info_types:
             previous_date = Resource._previous(info_types[info].keys(), date)
             return info_types[info][previous_date]
-            # if date in info_types[info]:
-            #     return info_types[info][date]
-            # else:
-            #     nearest_date = self._previous(info_types[info].keys(), date)
-            #     return info_types[info][nearest_date]
-                
+         
         else:
             logging.warning(f'No information about {info} is stored')
             return None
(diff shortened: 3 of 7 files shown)


collector/__init__.py
```diff
@@ -347,12 +347,13 @@ def _process_observation_data(self,
...
-                    for target in targets:
+                    for target in targets:                        
                         try:
                             target_name = target['group']['name']
-                            target_tag = TargetTag(target['tag']) if target_name == 'Base' else None
-                            target_designation = target['num'] if target_tag is not None and \
-                                                                  target_tag == TargetTag.MajorBody else target['des']
+                            if target_name == 'Base':
+                                target_tag = TargetTag(target['tag'])
+                                if target_tag is not TargetTag.Sidereal and target_tag is not None:
```
sraaphorst, line 355: You should switch the order of these two. Do the `None` check first always and for enums use `==`.
```
if target_tag is not None and target_tag != TargetTag.Sidereal:
```

collector/__init__.py
```diff
@@ -371,10 +372,10 @@ def _process_observation_data(self,
                 # Sky Conditions
                 conditions = get_conditions(raw_observation, label=False)
 
-                if conditions or conditions is None:
+                if not conditions or conditions is None:
```
sraaphorst, line 375: Again:
```
if conditions is None or not conditions:
```

greedy_max/schedule.py
```diff
@@ -164,11 +164,11 @@ def sky_constraints(self) -> None:
         Use the most restrictive value for each condition. 
         '''
         restrictive_iq = min([obs.sky_conditions.iq for obs in self.observations])
-        restrictive_bg = min([obs.sky_conditions.sb for obs in self.observations])
+        restrictive_sb = min([obs.sky_conditions.sb for obs in self.observations])
         restrictive_cc = min([obs.sky_conditions.cc for obs in self.observations])
         restrictive_wv = min([obs.sky_conditions.wv for obs in self.observations])
 
-        return SkyConditions(restrictive_iq,restrictive_bg,restrictive_cc,restrictive_wv)
+        return SkyConditions(restrictive_sb,restrictive_cc,restrictive_iq,restrictive_wv)
```
sraaphorst, line 171: Not to nitpick. but to get rid of warnings, we should put spaces after commas: it's a Python standard.

resource_mock/__init__.py
```diff
@@ -70,20 +65,20 @@ def _previous(items, pivot):
         return result
 
     def _excel_reader(self) -> NoReturn:
+
         sites = [site for site in Site]
-        workbook = load_workbook(filename=os.path.join(self.path, '2018B-2019A Telescope Schedules.xlsx'))
+        workbook = load_workbook(filename=f'{self.path}/2018B-2019A Telescope Schedules.xlsx')
```
sraaphorst, line 70: Just out of curiosity, why did you change this? We really should not be creating paths with strings, but should always use `os.path.join`, as it handles slashes in a platform independent way and manages extra slashes properly.

resource_mock/__init__.py
```diff
@@ -70,20 +65,20 @@ def _previous(items, pivot):
...
         for site in sites:
             sheet = workbook[site.name]
-            for row in sheet.iter_rows(min_row=2):
+            for row in sheet.iter_rows(min_row=2):                
                 date = row[0].value
-                self.instruments[site][date] = [c.value for c in row[3:]]
+                f2_filter = lambda x: 'Flamingos2' if x == 'F2' else x
```
sraaphorst, line 75: This isn't really a filter so much as a transformation. I would remove it and just do:
```
self.instruments[site][date] = ['Flamingos2' if c.value == 'F2' else c.value for c in row[3:]]
```

resource_mock/resources.py
```diff
@@ -23,6 +23,11 @@ def __init__(self,
                  ifus: Dict[str, Dict[str, str]]):
         self.fpu = fpu
         self.fpur = fpur
+        self.fpur[Site.GS].extend(['11013104','11013107','11020601','11022001',
+                                    '11023313','11023327','11023328','11023332',
+                                    '11023341','11023342','10000009','10005373',
+                                    '10005372','10005374','10005375','10005376',
+                                    '10005390'])
```
sraaphorst, line 30: This seems kind of hacky, no? Should this information be hard-coded? Just curious why it isn't getting passed in for `fpur` in the constructor.

resource_mock/resources.py
```diff
@@ -41,11 +46,17 @@ def is_instrument_available(self, site: Site, instrument: str) -> bool:
     def is_disperser_available(self, site: Site, disperser: str) -> bool:
         return disperser in self.gratings[site]
 
-    def is_mask_available(self, site: Site, fpu_mask: str) -> bool:
+    def is_mask_available(self,site: Site, fpu_mask: str) -> bool:
```
sraaphorst, line 49: Space after comma in `self`. 😉

resource_mock/resources.py
```diff
@@ -41,11 +46,17 @@ def is_instrument_available(self, site: Site, instrument: str) -> bool:
...
         elif '-' in fpu_mask:
-            barcode = Resources._decode_mask(fpu_mask)
-        return barcode and barcode in self.fpur[site]
-
+            barcode = self._decode_mask(fpu_mask)
+        
+        if barcode:
+            return barcode in self.fpur[site]
+        else:
+            return False # No mask in register
```
sraaphorst, line 62: All this should be collapsed into:
```
return barcode and barcode in self.fpur[site]
```

resource_mock/resources.py
```diff
@@ -41,11 +46,17 @@ def is_instrument_available(self, site: Site, instrument: str) -> bool:
...
         barcode = None
+        if fpu_mask == 'None':
+            return True
         if fpu_mask in self.fpu_to_barcode[site]:
             barcode = self.fpu_to_barcode[site][fpu_mask]
         elif '-' in fpu_mask:
-            barcode = Resources._decode_mask(fpu_mask)
-        return barcode and barcode in self.fpur[site]
-
+            barcode = self._decode_mask(fpu_mask)
```
sraaphorst, line 57: I'm not sure if some of these changes back came about because of problems with rebasing, but since `_decode_mask` can be written as a static method, we should be accessing it with `Resources._decode_mask` instead of `self._decode_mask`.

selector/__init__.py
```diff
@@ -479,8 +478,15 @@ def select(self, visits: List[Visit],
...
                     instruments_in_obs.append(comp_instrument)
                     valid_in_obs.append(comp_val)
+                    
                     vishours_of_obs.append(obs.visibility.hours[inight])
 
+                    # TODO: Calibrations should be consider? If this is not account
+                    # TODO: more visits are selected. 
+                    if visit.calibrations:
+                        for cal in visit.calibrations:
+                            vishours_of_obs.append(cal.visibility.hours[inight])
```
sraaphorst, line 488: I don't think you need the `if` statement here. If there are no `visit.calibrations`, then the loop:
```
for cal in visit.calibrations:
```
just won't do anything.

selector/__init__.py
```diff
@@ -494,32 +500,41 @@ def select(self, visits: List[Visit],
...
+                            if not resources.is_disperser_available(site, disperser):
+                                has_disperser = False
+                                break
+                        if has_disperser:
+                            for fpu in fpus_in_obs:
+                                if not resources.is_mask_available(site,fpu):
+                                    has_fpu = False
+                                    break
+                        if has_disperser and has_fpu:
+                            can_be_selected = True
```
sraaphorst, line 526: Good catch 😄. This looks like the right logic to me.

sraaphorst (changes requested):
Looks good for the most part. I asked some questions and added some points that should be cleaned up.

resource_mock/__init__.py
stroncod: I didn't. Might got lost in the rebase or something.

sraaphorst (approved):
Looks good!

resource_mock/__init__.py
sraaphorst: Yeah, `git rebase` has bit me in the ass more than a few times. 😒

PR #22, opened 2021-10-01
Multi-site implementation for the base structure

stroncod:


collector/__init__.py
```diff
@@ -74,7 +74,7 @@ def ot_timing_windows(starts: Iterable[int],
     return timing_windows
 
 
-def select_obsclass(classes):
+def select_obsclass(classes: List[str])-> str:
```
sraaphorst, line 77: We really should turn the obsclasses into a proper `Enum` since there are a limited number of them.
@bryanmiller Could you comment on what possible values there are for obsclasses?

collector/__init__.py
```diff
@@ -114,61 +114,51 @@ def __init__(self,
...
-
-        time_accounting = self._load_tas(path, site)
-        self._readzip(zip_path, xmlselect, site_name, tas=time_accounting, obsclasses=self.obs_classes)
+        
+        
+        for site in self.sites:
+            site_name = site.value
+
+            xmlselect = [site_name.upper() + '-' + sem + '-' + prog_type
+                        for sem in self.semesters for prog_type in self.program_types]
```
sraaphorst, line 136: `program_types` is another place where it seems we should use an `Enum` since there are only so many program types.
@bryanmiller Sorry for the repeated questions, but I assume these are types like Q, C, DD, LP, FT, etc?

collector/__init__.py
```diff
@@ -114,61 +114,51 @@ def __init__(self,
...
+
+            # TODO: We will have to modify in order for this code to be usable by other observatories.
+            zip_path = os.path.join(path,
+                                    f'{(self.time_range[0] - 1.0 * u.day).strftime("%Y%m%d")}{SITE_ZIP_EXTENSIONS[site]}')
+            logging.info(f'Retrieving program data from: {zip_path}.')
+
+            time_accounting = self._load_tas(path, site)
+            self._calculate_night_events(site)
+            self._readzip(zip_path, xmlselect, site, tas=time_accounting)
+            Collector.observation_num = 0
```
sraaphorst, line 159: I guess we set `Collector.observation_num` here to 0 because we are starting over for a different site?
Since this is never used outside of `Collector._process_observation_data`, I think we should eliminate it from being a member of `Collector` and just moving `observation_num` into `_process_observation_data`.

I think we can put it right after the new L132 and just write:
```
observation_num = 0
```

sraaphorst: Upon further investigation, maybe this is not so simple, but I would like to get this PR merged and then see if we can get rid of the `Collector.observation_num` variable entirely. I'll take a look at this after.

collector/__init__.py
```diff
@@ -251,6 +241,9 @@ def _instrument_setup(configuration: Dict[str, List[str]], instrument_name: str)
                 disperser = ulist[0]
             else:
                 instconfig[key] = ulist
+            
+            if 'Alopeke' in instrument_name:
+                instrument_name = 'Alopeke'
```
sraaphorst, line 246: Little confused by this. Is there any case where `Alopeke` would be IN the `instrument_name` but `instrument_name` would contain MORE than `Alopeke`?

Since my current codebase only runs for GS right now, I can't check it myself since Alopeke is only in GN.

greedy_max/schedule.py
```diff
@@ -163,10 +163,16 @@ def sky_constraints(self) -> None:
...
+        restrictive_sb = min([*[obs.sky_conditions.sb for obs in self.observations],
+                              *[c.sky_conditions.sb for c in self.calibrations]])
+        restrictive_cc = min([*[obs.sky_conditions.cc for obs in self.observations],
+                              *[c.sky_conditions.cc for c in self.calibrations]])
+
+        restrictive_iq = min([*[obs.sky_conditions.iq for obs in self.observations],
+                              *[c.sky_conditions.iq for c in self.calibrations]])
+
+        restrictive_wv = min([*[obs.sky_conditions.wv for obs in self.observations],
+                              *[c.sky_conditions.wv for c in self.calibrations]])
```
sraaphorst, line 175: Nice!

selector/__init__.py
```diff
@@ -383,33 +383,34 @@ def create_pool(self) -> Dict[Site, List[Visit]]:
         scheduling_groups = self.collector.scheduling_groups
         visits = {site: [] for site in self.sites}
 
+        print(self.sites)
```
sraaphorst, line 386: This should probably be deleted, or at least changed to a `logging.info` statement.

selector/__init__.py
```diff
@@ -383,33 +383,34 @@ def create_pool(self) -> Dict[Site, List[Visit]]:
...
 
-                instruments = [collected_observations[obs].instrument
+                instruments = [collected_observations[site][obs].instrument
                                for obs in obs_idxs]
 
                 wavelengths = set([wav for inst in instruments for wav in inst.wavelength()])
 
-                modes = [[collected_observations[obs].instrument.observation_mode()
+                modes = [[collected_observations[site][obs].instrument.observation_mode()
                           for obs in obs_idxs]]
```
sraaphorst, line 398: I didn't notice this before, but why do we make `modes` into a list containing a single element?
Why the doubly nested list?

sraaphorst (changes requested):
Good progress. This should put OR groups with multi-site observations in our minds and get us thinking as to how to handle them.

collector/__init__.py
sraaphorst: I'm a little confused, then. Why are we passing the `classes` as a `List[str]` and not a `List[ObservationClass]`?

collector/__init__.py
sraaphorst: Sounds good.

collector/__init__.py
sraaphorst: I figured that might be the case. I worked on the Alopeke implementation and I know that it can contain all kinds of additional information.

collector/__init__.py
```diff
@@ -441,36 +432,29 @@ def _process_observation_data(self,
...
-
-                self.observations.append(Observation(Collector.observation_num,
+                #self.priority.append(priority)
+
+                #logging.debug([oc  for oc in classes if oc != 'ACQ'][0] )
+                self.observations[site].append(Observation(Collector.observation_num,
                                                      obs_odb_id,
                                                      band,
-                                                     Category(obsclass.lower()),
+                                                     ObservationClass(obsclass.lower()),
```
sraaphorst, line 457: I see here. I think we can try to push this back so that `List[ObservationClass]` are being passed to the `Collector` instead of doing the transformation here, but for now, I'll mark this as resolved and investigate it later.

collector/__init__.py
sraaphorst: Never mind: I see further ahead what is going on.

selector/__init__.py
sraaphorst: Why shouldn't we fix it now, then?
I have noticed a lot of code currently goes unused, which is making me a bit nervous about correctness.

PR #23, opened 2021-10-06
Fixed bug with setting target name

bryanmiller:
_process_observation_data was overwriting the target_name with the target group name.  This fixes that.


sraaphorst (approved):
Thanks for catching this, Bryan.

PR #27, opened 2021-10-13
Sched 48: Review of TODO comments in source code

stroncod:
Some pending changes in the code. 
More information in this document: https://docs.google.com/document/d/1ZGEfh_OWWL1Q3_YRM93kJh2R8ST5gJKeGSDFSKvP34M/edit

XMLutils.py TODOs where excluded pending ODB extractor changes.

### DIFF
--- a/selector/__init__.py
+++ b/selector/__init__.py
@@ -22,7 +22,6 @@ from resource_mock.resources import Resources
 from astropy.coordinates import SkyCoord, Angle
 import astropy.units as u
 
-import multiprocessing
 import numpy as np
 from tqdm import tqdm
 
@@ -134,7 +133,7 @@ class Selector:
 
                 # File name
                 ephname = usite + '_' + des.replace(' ', '').replace('/', '') + '_' + \
-                          times[0].strftime('%Y%m%d_%H%M') + '-' + times[-1].strftime('%Y%m%d_%H%M') + '.eph'
+                    times[0].strftime('%Y%m%d_%H%M') + '-' + times[-1].strftime('%Y%m%d_%H%M') + '.eph'
 
                 try:
                     time, ra, dec = horizons.Coordinates(hzname, times[0], times[-1], step='1m', \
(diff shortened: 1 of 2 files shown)


common/structures/conditions.py
```diff
@@ -176,20 +176,14 @@ def __init__(self,
         self.wind_direction = wind_direction
         self.time_blocks = time_blocks
 
-    # TODO ERROR: This method needs to be fixed: speed is not defined unless the if condition holds.
     def get_wind_conditions(self, azimuth) -> np.ndarray:
-        if np.asarray(self.wind_speed).ndim == 0:
-            speed = np.full(len(azimuth), self.wind_speed.to(u.m / u.s).value) * u.m / u.s
-        
+
```
sraaphorst, line 180: I'm not totally clear now on what will happen if the `ndim` of `self.wind_speed` is `0`, or can this not happen?

common/structures/instrument.py
```diff
@@ -49,10 +83,10 @@ def gsaoi_mode():
         def nifs_mode():
             return 'ifu'
 
-        instrument_lookup = { 
+        instrument_lookup = {
                             'GMOS-S': gmos_mode,
                             'GMOS-N': gmos_mode,
```
sraaphorst, line 88: Since GMOS-N and GMOS-S have a lot of differences (e.g. different filters), we may have to separate these out at some point, but I don't think we need to worry about that right now.

common/structures/instrument.py
```diff
@@ -1,44 +1,78 @@
...
-            if any(['FPU_NONE' in fpu for fpu in self.configuration['fpu']]) \
-                    and any('IMAGING' in fpu for fpu in self.configuration['decker']):
-                return 'imaging'
+            if self.gmos_configuration is not None:
+                if any(['LONGSLIT' in fpu for fpu in self.gmos_configuration.fpu]):
+                    return 'longslit'
+
+                if any(['FPU_NONE' in fpu for fpu in self.gmos_configuration.fpu]) \
+                        and any('IMAGING' in fpu for fpu in self.decker):
+                    return 'imaging'
```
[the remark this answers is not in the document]
stroncod: I agree. I was in the understanding the FPU mask were only for GMOS. I think `gmos_configuration` is a bad name for that class, but I think that agrouppation is correct.

PR #28, opened 2021-11-16
Sched 64: Initial Process manager structures

stroncod:
First approach based on Sebastian's design document and a little of multiprocessing. Entirely separate package that does not impact anything in the current code.

### DIFF
--- a/process_manager/main.py
+++ b/process_manager/main.py
@@ -1,30 +1,48 @@
 from __init__ import ProcessManager
+import datetime
+
+from bin import SchedulerTask
+from scheduler import Scheduler
 
 if __name__ == "__main__":
     
+    # TODO: This is going to be moved to a config file
     config = {
-        'semestral_bins': {
+        'biannual_bins': {
             'amount': 1,
             'bin_size': 2,
-            'n_threads': 2
+            'n_threads': 2,
+            'length': datetime.timedelta(seconds=30),
+            'float_after': datetime.timedelta(seconds=10),
+            'start': datetime.datetime(2020, 1, 1, 0, 0, 0)
         },
-        'week_bins': {
+        'weekly_bins': {
             'amount': 1,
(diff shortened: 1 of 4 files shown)


stroncod:
The code right now does not test all the `add_task` method logic because a way to add task asynchronously is not yet implemented.

process_manager/main.py
```diff
@@ -0,0 +1,48 @@
+from __init__ import ProcessManager
+import datetime
+
+from bin import SchedulerTask
+from scheduler import Scheduler
+
+if __name__ == "__main__":
+    
+    # TODO: This is going to be moved to a config file
```
sraaphorst, line 9: Ah. Okay, now I see why you used `str` instead of `Enum`, but I think that we can just do an `Enum` lookup when we parse the config file:

```
class BinType(str, Enum):
    BIANNUAL_BINS = 'biannual_bins'
   ...

BinType('biannual_bins')
# returns <BinType.BIANNUAL_BINS: 'biannual_bins'>
```
Then we would get a `ValueError` automatically if the config file was wrong.

Same with the parameter values inside the bins: this will ensure to at least a greater degree that we don't have incorrect entries in the `dict`.

process_manager/scheduler.py
```diff
@@ -0,0 +1,8 @@
+import time
+
+
+class Scheduler():
+
+    def new_schedule(self, name):
+        time.sleep(5)
```
sraaphorst, line 7: In order to test properly, instead of a fixed sleep, we should probably make this a random time.

sraaphorst (changes requested):
Let's talk more about this tomorrow.

sraaphorst:
I'm assuming that Ricardo showed you his mock... I think there are a lot of really good ideas here that we should use:
https://github.com/rcardenes/sched_utils

PR #29, opened 2021-11-19
SCHED-71: Initial mini-model implementation.

sraaphorst:
Initial implementation of the mini-model, minus output and some internal data structures. (Will add these tomorrow.)

### DIFF
--- a/common/minimodel/__init__.py
+++ b/common/minimodel/__init__.py
@@ -449,11 +449,11 @@ class Program:
     band: Band
     thesis: bool
     mode: ProgramMode
-    too_type: Optional[TooType]
     start_time: datetime
     end_time: datetime
     allocated_time: Set[TimeAllocation]
     root_group: Group
+    too_type: Optional[TooType] = None
 
     def program_awarded(self) -> timedelta:
         return sum(t.program_awarded for t in self.allocated_time)
--- a/common/minimodel/__init__.py
+++ b/common/minimodel/__init__.py
@@ -12,6 +12,10 @@ from typing import List, Mapping, Optional, Set, Union
 
 
 class Site(str, Enum):
+    """
+    This will have to be customized by a given observatory if used independently
+    of Gemini.
+    """
     GN = 'Gemini North'
     GS = 'Gemini South'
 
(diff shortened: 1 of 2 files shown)


common/structures/conditions.py
```diff
@@ -196,3 +196,4 @@ def __init__(self,
                  wind: WindConditions):
         self.sky = sky
         self.wind = wind
+
```
sraaphorst, line 199: I have no idea why it is saying there is a diff between these two files.

common/minimodel/__init__.py
```diff
@@ -0,0 +1,474 @@
...
+    The wavelength must be specified in nanometers.
+    """
+    id: str
+    exec_time: timedelta
+    prog_time: timedelta
+    part_time: timedelta
+    observed: bool
+    qa_state: QAState
+    guide_state: bool
+    required_resources: Set[Resource]
```
stroncod, line 290: Won't each atom have the same Resource in a sequence?

common/minimodel/__init__.py
```diff
@@ -0,0 +1,474 @@
...
+    tag: TargetTag
+
+    def __post_init__(self):
+        super().__post_init__()
+
+
+@dataclass
+class Resource:
+    id: str
+    name: str
```
stroncod, line 267: If I recall the Selector would need more information (FPU, Gratings, etc). Is this going to be separate from the mini model ?

stroncod (approved):
I'm not sure if you want this to be merged or just reviewed but I approved just in case

common/minimodel/__init__.py
```diff
@@ -0,0 +1,474 @@
...
+        return {r for a in self.sequence for r in a.required_resources}
+
+    def wavelengths(self) -> Set[float]:
+        return {c.wavelength for c in self.sequence}
+
+    def constraints(self) -> Set[Constraints]:
+        return {self.constraints}
+
+
+# Since Python doesn't allow classes to self-reference, we have to make a basic group
```
stroncod, line 365: I understand this is an issue with using type annotations??? Using literal string might solve that.

stroncod (approved):

sraaphorst:
I'm merging this, although it is a WIP and we will need to adapt it as we go. Outstanding tasks currently include:

- Addition of the instrument configuration at the atom level.
- Decisions about where time delays fit (i.e. not in Group, but possibly in Observation) for cadences based on Bryan's feedback. Waiting for some examples to make sure that we can model cadences appropriately.

PR #31, opened 2021-11-26
Sched 70: Process Manager 

stroncod:
Just for reviewing the code. Not to be merge!

### DIFF
--- a/process_manager/__init__.py
+++ b/process_manager/__init__.py
@@ -1,15 +1,16 @@
 from datetime import datetime, timedelta
-import multiprocessing as mp
+
 from bin import SchedulingBin, RealTimeSchedulingBin, SchedulerTask, BinType
-from heapq import heappush, nsmallest, heappop
 import asyncio
 import signal
 from typing import  List, Tuple, Optional
+from provider import TaskProvider
+import logging
+
 
 class ProcessManager:
-    def __init__(self, config: dict = None) -> None:
+    def __init__(self, config: List = None, period: int = 5) -> None:
         
-        # TODO: if these are fix should we use Enum?
         self.bins = {BinType.REALTIME: [],
                      BinType.STANDARD: []}
         if config is not None:
@@ -27,6 +28,10 @@ class ProcessManager:
                                                                  cfg.bin_size))
                 else:
                     raise ValueError('BinType unknown')
+        self.period = period
+        self.queue = asyncio.Queue()
+        self.task_provider = TaskProvider()
+
 
     def new_bin(self,
                 bin_type: str,
@@ -35,6 +40,9 @@ class ProcessManager:
                 length: timedelta,
                 number_threads: int,
                 bin_size: int) -> None:
+        """
+        Adds a new bin to the process manager
+        """
         if bin_type == 'realtime':
             if len(self.bins['realtime']) > 1:
                 raise ValueError('Only one bin for realtime mode')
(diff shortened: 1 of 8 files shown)


process_manager/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+                                                                 cfg.number_threads,
+                                                                 cfg.bin_size))
+                else:
+                    raise ValueError('BinType unknown')
+        self.period = period
+        self.queue = asyncio.Queue()
+        self.task_provider = TaskProvider()
+
+
+    def new_bin(self,
```
sraaphorst, line 36: I don't see this being called anywhere. Are we using it?

process_manager/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+                start: datetime,
+                float_after: timedelta,
+                length: timedelta,
+                number_threads: int,
+                bin_size: int) -> None:
+        """
+        Adds a new bin to the process manager
+        """
+        if bin_type == 'realtime':
+            if len(self.bins['realtime']) > 1:
```
sraaphorst, line 47: We should switch these from `str` to `BinType`. Same below for `realtime`.

process_manager/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+                raise ValueError('Only one bin for realtime mode')
+            self.bins['realtime'].append(RealTimeSchedulingBin(start, float_after, length, number_threads, bin_size))
+       
+        else:
+            if bin_type not in self.bins:
```
sraaphorst, line 52: If we're using an `Enum` for `BinType`, then this check shouldn't really be necessary, but I guess without strong typing in Python, maybe it's a good idea all the same.

process_manager/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+                raise ValueError('Bin type not supported')
+            self.bins[bin_type].append(SchedulingBin(start, float_after, length, number_threads, bin_size))
+    
+    async def run(self) -> None:
+        """
+        Main driver for the process manager.
+
+        This runs in a loop asyncrhonously adding tasks to each bin accordinly.
```
sraaphorst, line 60: Not to be nitpicky, but just a couple small typos: asynchronously and accordingly.

process_manager/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+        ptask = asyncio.create_task(self.task_provider.producer(self.queue, 1))
+        ctask = asyncio.create_task(self.task_provider.consumer(self.queue))
+
+        loop = asyncio.get_event_loop()
+        loop.add_signal_handler(signal.SIGINT, shutdown, ptask, ctask)
+
+        while not done.is_set():
+            
+            task = await ctask
+            print(task)
```
sraaphorst, line 86: Let's switch to logging here, and it might be a good idea to use more of it, because this will likely turn out to be iffy to debug, so all the information we can get will be handy.

process_manager/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+            if task is not None:
+                if task.is_realtime:
+
+                    logging.info(f"Real Time Scheduling a job for {task}")
+                    self.bins[BinType.REALTIME].schedule_with_runner(task)
+                    await asyncio.sleep(self.period)
```
sraaphorst, line 92: In order to have the dummy tasks actually "pretend" to be doing something, we should do the sleeping in the tasks themselves in a kind of `execute` method, but call the `await` on that method here and then log something for the results to indicate that this is where we will handle the results.

process_manager/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+        loop.add_signal_handler(signal.SIGINT, shutdown, ptask, ctask)
+
+        while not done.is_set():
+            
+            task = await ctask
+            print(task)
+            if task is not None:
+                if task.is_realtime:
+
+                    logging.info(f"Real Time Scheduling a job for {task}")
```
sraaphorst, line 90: 👍

process_manager/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+                    self.bins[BinType.REALTIME].schedule_with_runner(task)
+                    await asyncio.sleep(self.period)
+                else:
+                    logging.info(f"Standard Scheduling a job for {task}")
+                    for bin in self.bins[BinType.STANDARD]:
+                        if bin.start <= task.start_time and bin.start + bin.length < task.end_time:
+                            bin.schedule_with_runner(task)
+                            break
```
sraaphorst, line 98: As a sanity check, if a task does not fall in an existing bin, we should log an error.

process_manager/bin.py
```diff
@@ -0,0 +1,129 @@
...
+        self.target = target
+        self.process = None
+    
+    def __repr__(self):
+        return f"{self.target} <- {{prio={self.priority}, timeout={self.timeout}}}"
+
+
+class BinType(Enum):
+    REALTIME = 'realtime'
+    STANDARD = 'standard'
```
sraaphorst, line 41: 👍

process_manager/bin.py
```diff
@@ -0,0 +1,129 @@
...
+        prun = PriorityRunner(self.bin_size)
+        prun.add_done_callback(self._schedule_pending)
+        self.runner = prun
+        self.pending_tasks = []
+        self.accepting = True
+        self.priority_queue = []
+        self.running_tasks = []
+
+    def float_bin(self):
+        self.start += self.float_after
```
sraaphorst, line 76: A little bit of thought might still be needed here on the design side of things. Since science asked that we float daily and week bins constantly, they will require special handling, but this is just a placeholder more to me to clarify that and for us to make sure it is implemented down the road. Right now, this is not a high priority modification.

Maybe we can add a:
```
# TODO: Float daily and weekly bins regularly.
```
just so that it's in the code base.

process_manager/bin.py
```diff
@@ -0,0 +1,129 @@
...
+        # check if task are still on bounds #
+        total_time = self.start + self.length
+
+        # check on pending tasks #
+        for task in self.pending_tasks:
+            if task.end_time > total_time:
+                # remove from queue 
+                pass
```
sraaphorst, line 84: ? I guess this a placeholder for now? Are there any challenges to remove them from the queue now?

process_manager/bin.py
```diff
@@ -0,0 +1,129 @@
...
+    def shutdown(self):
+        """
+        Attempt to "gracefully" terminate all running tasks.
+        """
+        self.accepting = False
+        self.runner.terminate_all()
+        self.running_tasks = []
+
+class RealTimeSchedulingBin(SchedulingBin):
+    def __init__(self, start, float_after, length) -> None:
```
sraaphorst, line 128: Some type hints here would be good, I think. Same with a couple other methods / constructors above.

process_manager/process.py
```diff
@@ -0,0 +1,141 @@
...
+    if asyncio.iscoroutine(res):
+        await res
+
+class ProcessTask:
+    '''
+    A class that allows an async process fine control over a
+    multiprocessing.Process
+
+    Initialize by passing an instance of Process, that must not
+    be running.
```
sraaphorst, line 32: Can we check this to make sure?

process_manager/process.py
```diff
@@ -0,0 +1,141 @@
...
+    def terminate(self, result=Result.TERMINATED):
+        """
+        Cancel the task waiting for the process to finish. As a side
+        effect, this will terminate the process itself.
+
+        Sets the result to the specified value.
+        """
+        if self.process.is_alive():
+            if self.result is None:
+                self.result = result
```
sraaphorst, line 82: Not quite clear why we don't just set `self.result = result` and do the `None check.

process_manager/provider.py
```diff
@@ -0,0 +1,55 @@
...
+from dataclasses import dataclass
+
+
+@dataclass
+class TaskGenerator:
+
+    def generator(self):
+        while True:
+            task = SchedulerTask(datetime(2020, 1, 1, 0, 0, 0),
+                                 datetime(2020, 1, 1, 0, 0, 0),
```
sraaphorst, line 17: Why the same start and end times?

process_manager/provider.py
```diff
@@ -0,0 +1,55 @@
...
+import time
+from datetime import datetime
+from random import randint
+from scheduler import Scheduler
+import logging
+from dataclasses import dataclass
+
+
+@dataclass
+class TaskGenerator:
```
sraaphorst, line 12: This looks good, but it seems like it will never generate real time tasks?

sraaphorst (approved):
This looks like it's shaping up pretty well. Just a couple small changes to make, and we need to look at `ProcessManager.new_bin` to see if / why we still need this since it's not being called.

I'm going to mark it approved and we'll revisit it later, but if you could make the small changes I detail below, that would be great! 👍

sraaphorst:
Without running this myself, it's kind of hard to get a sense that it is doing what we need it to do, but it's a good starting place for us to be in, I think.

sraaphorst:
Are we still going to be using this or should this be closed?

PR #33, opened 2021-12-01
Example script for manipulating ODB extractor json and defining atoms

bryanmiller:


stroncod (approved):

odb_extractor_atoms.py
```diff
@@ -0,0 +1,627 @@
...
+        else:
+            coadds = int(step['observe:coadds'])
+
+        disperser = step['instrument:disperser']
+
+        # Offsets
+        if 'telescope:p' in keys:
+            p = float(step['telescope:p'])
+        else:
+            p = 0.0
```
sraaphorst, line 56: Just curious why p offsets are not held in a list, but q offsets are?

odb_extractor_atoms.py
```diff
@@ -0,0 +1,627 @@
...
+                    nabba = 3
+                    nextatom = True
+                    print('Atom for ABBA')
+        else:
+            nabba -= 1
+        
+        if nextatom:
+            natom += 1
+            atoms.append({'id': natom, 'exec_time': 0.0, 'prog_time': 0.0, 'part_time': 0.0, 'filter': filter_name,
+                          'fpu': fpu, 'disperser': disperser, 'wavelength': wavelength, 'guiding': guiding})
```
sraaphorst, line 92: Seems to me that we should have an `Atom` dataclass that has these fields instead of a dict? And `filter`, `fpu`, and `disperser` should all be instances of `Resource` in the mini-model? Does anyone have any objections if I change this for consistency?

odb_extractor_atoms.py
```diff
@@ -0,0 +1,627 @@
...
+def printseq(sequence, comment='', csv=False, path=''):
+    '''Print basic configuration information about a sequence, with an option to write to a csv file'''
+
+    atom = '1'
+    if csv and path != '':
+        obsid = sequence[0]['ocs:observationId']
+        filename = os.path.join(path, obsid + '_seq.csv')
+        f = open(filename, 'w')
+        print('{},{}'.format('comment', comment), file=f)
+        print('{},{},{},{},{},{},{},{},{},{},{},{}'.format('datalab', 'class', 'exptime', 'coadds', 'inst', 'fpu',
```
sraaphorst, line 120: In Python 3, this could be written:
```
print(f'{datalab}{class}{exptime}{coadds}{inst}{fpu}{filter_name}{disperser}{wavelength}{p}{q}{atom}', file=f)
```

odb_extractor_atoms.py
```diff
@@ -0,0 +1,627 @@
...
+            # convert wavelength to microns
+#             wavelength = '{:5.3f}'.format(float(wavelength) / 1000.)
+        else:
+            coadds = step['observe:coadds']
+        disperser = step['instrument:disperser']
+        if 'telescope:p' in step.keys():
+            p = step['telescope:p']
+        else:
+            p = '0.0'
+        if 'telescope:q' in step.keys():
```
sraaphorst, line 146: These magic values should be defined elsewhere once and referenced.

odb_extractor_atoms.py
```diff
@@ -0,0 +1,627 @@
...
+               'disperser', 'wavelength', 'p', 'q', 'atom']
+    
+    row = 2
+    for ii, col in enumerate(columns):
+        _ = ws.cell(column=ii+1, row=row, value="{0}".format(col))
+    row += 1
+        
+#     print('{},{}'.format('comment', comment), file=f)
+#     print('{},{},{},{},{},{},{},{},{},{},{}'.format('datalab', 'class', 'exptime', 'coadds', 'inst', 'fpu', 
+#                                                            'disperser', 'wavelength', 'p', 'q', 'atom'), file=f)
```
sraaphorst, line 185: Remove?

odb_extractor_atoms.py
```diff
@@ -0,0 +1,627 @@
...
+
+    for step in list(sequence):
+        data = []
+        data.append(step['observe:dataLabel'])
+        data.append(step['observe:class'])
+        inst = step['instrument:instrument']
+        data.append(inst)
+    #     print(inst, fpuinst[inst])
```
sraaphorst, line 193: Remove?

odb_extractor_atoms.py
```diff
@@ -0,0 +1,627 @@
...
+        data.append(float(step['observe:exposureTime']))
+        if 'GMOS' in inst:
+            coadds = '1'
+            # convert wavelength to microns
+#             wavelength = '{:5.3f}'.format(float(wavelength) / 1000.)
```
sraaphorst, line 198: Unsure why we're formatting `wavelength` as `str` instead of `float`?

odb_extractor_atoms.py
```diff
@@ -0,0 +1,627 @@
...
+
+    filename = os.path.join(path, file)
+    f = open(filename, 'r')
+    
+    sequence = {}
+    
+    # Read and parse csv file: first line is a comment, second has column headings
+    nline = 0
+    for line in f:
+#         line = line.rstrip('\n')
```
sraaphorst, line 242: Remove?

odb_extractor_atoms.py
```diff
@@ -0,0 +1,627 @@
+#!/usr/bin/env python
```
sraaphorst, line 1: For example scripts like this, maybe we should put them in their own package for reference and to not muddy the waters with the actual Scheduler code.

odb_extractor_atoms.py
bryanmiller: Sure. I'm just using a dictionary for convenience, not to suggest that is the way to model atoms.

odb_extractor_atoms.py
bryanmiller: It should be a float in the scheduler. I converted it to a string since in this case I'm just using it in a print statement.

odb_extractor_atoms.py
bryanmiller: That would be fine. Maybe in my old 'greedy_max' repo?

odb_extractor_atoms.py
bryanmiller: I was only only concerned with identifying ABBA patterns which just offset in q. The qoffsets list was an early idea and is not actually used. In my current version I use the qoffsets variable for something else.

PR #34, opened 2021-12-03
Sched 74: Process manager new structure

stroncod:
This work independently from the rest of the code. A first approach to the process manager for both standard and realtime uses.

### DIFF
--- a/process_manager/main.py
+++ b/process_manager/main.py
@@ -5,13 +5,16 @@ import asyncio
 import signal
 from datetime import datetime, timedelta
 from random import random, randint
-from task import DEFAULT_PERIOD, DEFAULT_POOL_SIZE, DEFAULT_TIMEOUT, SchedulerTask
+from task import DEFAULT_TIMEOUT, SchedulerTask, TaskType
 from scheduler import Scheduler
 from manager import ProcessManager
 
 # To mock dates
 START_DATE = datetime(2020, 1, 1, 0, 0)
 END_DATE = datetime(2021, 1, 1, 0, 0)
+
+DEFAULT_PERIOD = 5
+DEFAULT_POOL_SIZE = 10
 
 def set_logging():
     logging.basicConfig(level=logging.INFO,
@@ -30,8 +33,8 @@ def parse_cmdline():
 
     return parser.parse_args()
 
-# TODO: Period value in the args should be a random value between min and max if we want 
-# to mimic a subscription scenerio 
+# TODO: Period value in the args should be a random value between min and max if we want
+# to mimic a subscription scenerio
 def get_new_task(timeout: int):
     random_date = random() * (END_DATE - START_DATE) + START_DATE
     return SchedulerTask(random_date,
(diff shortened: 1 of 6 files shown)


process_manager/manager.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def schedule_with_runner(self, task: SchedulerTask, mode: str):
+        """
+        Schedule a task with the corresponding runner for the given mode.
+        """
+        # TODO: Probably good for enums but right now the original input is a string
+        # so it seems unnecessary to use enums?
+        if mode == 'realtime':
+            return self.realtime_runner.schedule(Process(target=task.target), task.timeout)
+        elif mode == 'standard':
```
sraaphorst, line 22: Not sure why using `str` here instead of `Enum`. We should probably also be raising a `ValueError` if values don't match, but converting from `str` to `Enum` will likely handle this.

process_manager/runner.py
```diff
@@ -0,0 +1,124 @@
...
+
+from process import ProcessTask, Result
+from multiprocessing import Process
+
+job_counter = count()
+
+@dataclasses.dataclass(order=True)
+class Job:
+    process: ProcessTask = dataclasses.field(compare=False)
+    sequence: int = dataclasses.field(compare=False,
```
sraaphorst, line 15: Maybe this is something I don't know about `dataclass`, but if you mark both fields as `compare=False`, what does `order` compare on?

process_manager/runner.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        if len(self.jobs) == self.max_jobs:
+            self.evict()
+        if len(self.jobs) < self.max_jobs:
+            self.jobs.append(self._run_job(process, timeout))
+            return True
+        else:
+            return False
+    
+    def terminate_all(self) -> None:
```
sraaphorst, line 117: These should probably be marked as `NoReturn`, I would think, but I don't suppose it really matters that much.

process_manager/task.py
```diff
@@ -0,0 +1,23 @@
...
+DEFAULT_PERIOD = 5
+DEFAULT_POOL_SIZE = 10
+
+
+class SchedulerTask:
+    def __init__(self,
+                 start_time: datetime,
+                 end_time: datetime,
+                 target: callable,
+                 timeout: int = 10) -> None:
```
sraaphorst, line 13: `= DEFAULT_TIMEOUT`?

process_manager/main.py
```diff
@@ -0,0 +1,66 @@
+
+import argparse
+import logging
+import asyncio
+import signal
+from datetime import datetime, timedelta
+from random import random, randint
+from task import DEFAULT_PERIOD, DEFAULT_POOL_SIZE, DEFAULT_TIMEOUT, SchedulerTask
```
sraaphorst, line 8: If we're not using these in `task` but using them here, why not just define them here?

sraaphorst (approved):
Some small changes and it's hard to follow without actually running this, but I'm approving it so we can maybe see it in action today / tomorrow.

PR #38, opened 2021-12-10
SCHED-71: Mini model AND OR group refinements.

sraaphorst:
AND groups (`AndGroup`) now explicitly require `num_to_observe` to be EQUAL to the length of its children.

OR groups (`OrGroup`) now explicitly require `num_to_observe` to be LESS THAN the length of its children.

Both groups no longer allow for `num_to_observe` to be 0.

### DIFF
--- a/common/minimodel/__init__.py
+++ b/common/minimodel/__init__.py
@@ -9,7 +9,7 @@ from datetime import datetime, timedelta
 from enum import Enum, IntEnum, auto
 import numpy.typing as npt
 from pytz import timezone, UnknownTimeZoneError
-from typing import ClassVar, List, Mapping, Optional, Set, Tuple, Union
+from typing import ClassVar, List, Mapping, Optional, Set, Union
 
 
 class SiteInformation:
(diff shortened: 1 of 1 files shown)


stroncod (approved):

common/minimodel/__init__.py
```diff
@@ -469,11 +469,13 @@ class NodeGroup(Group, ABC):
...
-        if self.number_to_observe < 0 or self.number_to_observe > len(self.children):
-            raise ValueError(f'Group {self.group_name} has {len(self.children)}, requires {self.number_to_observe}.')
+        if self.number_to_observe <= 0:
+            msg = f'Group {self.group_name} specifies non-positive {self.number_to_observe} children to be observed.'
+            logging.error(msg)
+            raise ValueError(msg)
 
     def required_resources(self) -> Set[Resource]:
-        return {r for c in self.children for r in c.required_resources()}
+        return {r for c in self.children for r in c.reqfsssuired_resources()}
```
sraaphorst, line 478: Well, I did something unusual here... I'll fix this today.

PR #39, opened 2021-12-10
Sched 76: Api implementation

stroncod:
PR just for review, not to be merged.

Everything with three dots is something was not clear to me where should I get the info from the JSON file. 
That includes both start and end times for Program, right know initialized with `None`


test_provider.py
```diff
@@ -0,0 +1,149 @@
...
+    def __init__(self, path):
+        self.path = path
+
+    def load_program(self, path: str) -> dict:
+        with open(path, 'r') as f:
+            return json.loads(f.read())
+    
+    @staticmethod
+    def parse_magnitude(json: dict) -> Magnitude:
+        ...
```
sraaphorst, line 17: Do you need help with this one? We can try to work on it together.

test_provider.py
```diff
@@ -0,0 +1,149 @@
...
+        
+        constraints.cc = json['cc']
+        constraints.iq = json['iq']
+        constraints.sb = json['sb']
+        constraints.wv = json['wv']
+        constraints.elevation_type = json['elevationConstraintType']
+        constraints.elevation_min = json['elevationConstraintMin']
+        constraints.elevation_max = json['elevationConstraintMax']
+        constraints.timing_windows = JsonProvider.parse_timing_window(json['timingWindows']) # this would need parsing 
+        constraints.strehl = ...
```
sraaphorst, line 42: Don't worry about strehl for now. This is fine. It'll come into play later.

test_provider.py
```diff
@@ -0,0 +1,149 @@
...
+        constraints = Constraints()
+        
+        constraints.cc = json['cc']
+        constraints.iq = json['iq']
+        constraints.sb = json['sb']
+        constraints.wv = json['wv']
+        constraints.elevation_type = json['elevationConstraintType']
+        constraints.elevation_min = json['elevationConstraintMin']
+        constraints.elevation_max = json['elevationConstraintMax']
+        constraints.timing_windows = JsonProvider.parse_timing_window(json['timingWindows']) # this would need parsing
```
sraaphorst, line 41: Not quite clear what you mean by the comment here.

test_provider.py
```diff
@@ -0,0 +1,149 @@
...
+
+        atom.id = ...
+        atom.exec_time = ...
+        atom.prog_time = ...
+        atom.part_time = ...
+        atom.observed = json['metadata:complete']
+        atom.qa_state = ...
+        atom.guide_state = ...
+        atom.required_resources = ...
+        atom.wavelength = ...
```
sraaphorst, line 56: While `Atom` is not `frozen` because we will be likely adjusting the times, I think it would still be good to parse out:
```
id = ...
exec_time = ...
...
return Atom(id, exec_time, ...)
```

test_provider.py
```diff
@@ -0,0 +1,149 @@
...
+        return atom
+
+    @staticmethod
+    def parse_sidereal_target(json: dict) -> SiderealTarget:
+        sidereal = SiderealTarget()
```
sraaphorst, line 61: Same. We don't know which classes will end up `frozen` and I think that `SiderealTarget` is probably a good candidate, so extract the contents and then create a `SiderealTarget` from them.

test_provider.py
```diff
@@ -0,0 +1,149 @@
...
+
+    @staticmethod
+    def parse_nonsidereal_target(json: dict) -> NonsiderealTarget:
+        nonsidereal = NonsiderealTarget()
+        nonsidereal.des = json['base']['des']
+        nonsidereal.tag = json['base']['tag']
+        nonsidereal.ra = ...
+        nonsidereal.dec = ...
+
+        return nonsidereal
```
sraaphorst, line 77: Same. We'll need logic to call the HORIZONS lookup and populate `ra` and `dec` but let's not worry about this here: this can be done by another independent process in the Collector so that it is handled for all `ProgramProvider`s.

test_provider.py
```diff
@@ -0,0 +1,149 @@
...
+        obs.program_used = ...
+        obs.partner_used = ...
+        targets_in_json = [json[key] for key in json.keys() if key.startswith('TELESCOPE_TARGETENV')]
+        targets = [JsonProvider.parse_sidereal_target(target) if target['base']['tag'] == 'sidereal' else JsonProvider.parse_sidereal_target(target) for target in targets_in_json]
+        obs.targets = targets
+        obs.guide_stars = ...
+        obs.sequence = [JsonProvider.parse_atom(seq) for seq in json['sequence']]
+        find_constraints = [json[key] for key in json.keys() if key.startswith('SCHEDULING_CONDITIONSt')]
+        obs.constraints = JsonProvider.parse_constraints(find_constraints[0])
+        obs.too_type = ...
```
sraaphorst, line 103: Same.

test_provider.py
```diff
@@ -0,0 +1,149 @@
...
+        return ta
+
+
+    @staticmethod
+    def parse_or_group(json: dict) -> OrGroup:
+        ...
+
+    @staticmethod
+    def parse_and_group(json: dict) -> AndGroup:
+        ...
```
sraaphorst, line 122: These should be pretty easy for the ODB, I think. We don't need to implement `parse_or_group` and `parse_and_group` will just correspond to scheduling groups and possibly organizational folders.

We will need special logic in `Observation` and `Program` to make sure that the root group and the terminal node groups are handled properly, but I don't think this should be too hard.

Let me know if you're having problems with this and we can discuss.

PR #41, opened 2021-12-17
Add program processing, class/qastates for atoms

bryanmiller:
Here is a new version with the following changes:
- can process an entire program at once, working down through the groups and observations to define atoms (prog_proc)
- calculates obsclass and qastate for each atom and propagates up to observation. These should agree with the values from the ODB given at the observation level. This won't be exact yet. A QA state of 'NONE' means not observed.

This is meant to help me review the results of the atom algorithm more broadly and it shows how atom properties are calculated from the included sequence steps and the propagated to the observation level.

This currently works on the observations without much selection or other modification. For the validation mode we will want to 'unobserve' most of the executed observation (e.g. unset qastates) and ignore acquisitions.

### DIFF
--- a/odb_extractor_atoms.py
+++ b/odb_extractor_atoms.py
@@ -8,6 +8,7 @@
 import os
 import json
 import gzip
+import numpy as np
 
 from openpyxl import Workbook
 from openpyxl import load_workbook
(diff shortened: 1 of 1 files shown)


stroncod (approved):

odb_extractor_atoms.py
```diff
@@ -16,98 +17,248 @@
...
+    # Precedence order for observation classes.
+    qastate_order = ['NONE', 'UNDEFINED', 'FAIL', 'USABLE', 'PASS']
+
+    # Set the qastate for the entire observation based on precedence
+    for state in qastate_order:
+        if state in states:
+            qastate = state
+            break
+
+    return qastate
```
sraaphorst, line 36: Just a note that by exploiting ordering in `IntEnum`, we can greatly simplify this by performing a small tweak in the the mini-model (which is similar to what I did with `ObservationState`). Right now, we have:
```
class QAState(Enum):
    UNKNOWN = auto()
    USABLE = auto()
    PASS = auto()
    FAIL = auto()
```

By changing this to:
```
class QAState(IntEnum):
    NONE = auto()
    UNDEFINED = auto()
    FAIL = auto()
    USABLE = auto()
    PASS = auto()
```

Then this can be collapsed down to:

```
def select_qastate(states: List[QAState]) -> Optional[QAState]:
    return None if len(states) == 0 else min(states)
```

odb_extractor_atoms.py
```diff
@@ -16,98 +17,248 @@
...
+    # Precedence order for observation classes.
+    obsclass_order = ['SCIENCE', 'PROGCAL', 'PARTNERCAL', 'ACQ', 'ACQCAL', 'DAYCAL']
+
+    # Set the obsclass for the entire observation based on obsclass precedence
+    for oclass in obsclass_order:
+        if oclass in classes:
+            obsclass = oclass
+            break
+
+    return obsclass
```
sraaphorst, line 57: Same as mentioned in the above.

PR #42, opened 2021-12-17
Sched 76: API Implementation

stroncod:


common/minimodel/__init__.py
```diff
@@ -366,15 +367,15 @@ class Atom:
...
-
+    NEW = 'NEW'
+    INCLUDED = 'INCLUDED'
+    PROPOSED = 'PROPOSED'
+    APPROVED = 'APPROVED'
+    FOR_REVIEW = 'FOR_REVIEW'
+    READY = 'READY'
+    ONGOING = 'ONGOING'
+    OBSERVED = 'OBSERVED'
+    INACTIVE = 'INACTIVE'
```
sraaphorst, line 378: I would prefer not to do this, because the ordering of these makes things much simpler.
We can do lookups by leaving things as is and doing:
```
ObservationStatus['NEW']
```
which returns:
```
<ObservationStatus.NEW: 1>
```
if we stick with `auto()`.

These are considered to have precedence, so we can compare them and in a list of them get the minimum by doing:

```
def min_obs_status(obs_statuses: List[ObservationStatus]): Optional[ObservationStatus] ->
    return min(obs_statuses, default=None)
```

odb_extractor_atoms.py
```diff
@@ -346,7 +348,7 @@ def seqxlsx(sequence, comment='', path=''):
         if 'GMOS' in inst:
             coadds = '1'
             # convert wavelength to microns
-#             wavelength = '{:5.3f}'.format(float(wavelength) / 1000.)
+    #             wavelength = '{:5.3f}'.format(float(wavelength) / 1000.)
```
sraaphorst, line 351: As a note, `format` is pretty outdated. We should be using string interpolation. (This is a general comment and not to anyone in specific.)

```
print(f'{wavelength:5.3f}')
```
or even:
```
print(f'{float(wavelength)/1000:5.3f}')
```

test_provider.py
```diff
@@ -0,0 +1,310 @@
...
+    def parse_guide_star(json: dict) -> Resource:
+
+        # TODO: Maybe a GuideStart class should be in place to handle this better
+        res = Resource(json['name'], json['tag'])
+        return res
+    
+    @staticmethod
+    def parse_root_group(json: dict) -> OrGroup:
+        # Find nested OR groups/AND groups
+        groups = [JsonProvider.parse_and_group(json[key]) for key in json.keys() if key.startswith('GROUP_GROUP_SCHEDULING')]
```
sraaphorst, line 271: Just a note: we may have to handle `ORGANIZATIONAL_FOLDER` as well, which I think would be the same as a scheduling group, but I'll have to test.

test_provider.py
```diff
@@ -0,0 +1,310 @@
...
+
+        number_to_observe = len(observations)
+        delay_max, delay_min = 0, 0 # TODO: What are these?
+        or_group = OrGroup(json['key'], json['name'], number_to_observe, delay_min, delay_max, observations)
+        return or_group
+
+    @staticmethod
+    def parse_guide_star(json: dict) -> Resource:
+
+        # TODO: Maybe a GuideStart class should be in place to handle this better
```
sraaphorst, line 264: I think a `GuideStar` should just be a `SiderealTarget`.

test_provider.py
```diff
@@ -0,0 +1,310 @@
...
+        with open(path, 'r') as f:
+            return json.loads(f.read())
+    
+    @staticmethod
+    def sort_observations(obs: List[Observation]) -> List[Observation]:
+        return sorted(obs, key=lambda x: x.order)
+
+    @staticmethod
+    def parse_magnitude(json: dict) -> Magnitude:
+        band = MagnitudeBand(json['name'], None, None)
```
sraaphorst, line 36: For `MagnitudeBand`, we should never have to create these. We should just be selecting the one with the given `name` from `MagnitudeBands`:
```
band =MagnitudeBands[json['name']]
```
Then create the `Magnitude` like this:
```
mag = Magnitude(band, json['value'], None)
```
We probably do want to assert that the `system` matches the `system` in the `MagnitudeBand` that we retrieve and if it doesn't, log and raise a `ValueError`.

sraaphorst: I added comments to the mini-model to make this more explicit, and will add some more comments because perhaps my design choices might be a bit unclear.

common/minimodel/__init__.py
```diff
@@ -389,9 +390,9 @@ class TooType(IntEnum):
 
 
 class SetupTimeType(Enum):
-    FULL = auto()
-    REACQ = auto()
-    NONE = auto()
+    FULL = 'FULL'
+    REACQ = 'REACQ'
+    NONE = 'NONE'
```
sraaphorst, line 395: I don't think it is ever needed to have an `Enum` with its names equal to its entries, as mentioned above.

common/minimodel/__init__.py
```diff
@@ -535,9 +537,9 @@ class ProgramMode(Enum):
     * Classical
     * Priority Visitor (hybrid mode between queue and classical)
     """
-    QUEUE = auto()
-    CLASSICAL = auto()
-    PV = auto()
+    QUEUE = 'Queue'
+    CLASSICAL = 'Classical'
+    PV = 'PV'
```
sraaphorst, line 542: Again. We can do lookups by:
```
value = ProgramMode[key.upper()]
```
and leave these as `auto`.

odb_extractor_atoms.py
```diff
@@ -286,7 +288,7 @@ def printseq(sequence, comment='', csv=False, path=''):
         if 'GMOS' in inst:
             coadds = '1'
             # convert wavelength to microns
-#             wavelength = '{:5.3f}'.format(float(wavelength) / 1000.)
+        #     wavelength = '{:5.3f}'.format(float(wavelength) / 1000.)
```
sraaphorst, line 291: I'm not quite clear why we're converting `wavelength` to  a float and then back to a `str`.
Do we need to do this for precision?

We should avoid using `.format` when possible as it is outdated and string interpolation is the preferred `f`-strings:
```
wavelength = f'{float(wavelength) /  1000:5.3f}'
```
I'm also unclear as to why we want the `5`, in the `:5.3f`, which doesn't seem to affect anything from what I can tell.

If we want to convert to microns, though, we could probably just do:
```
# Read wavelength in as nanometers and convert to micron
wavelength = f'{float(step['instrument:observingWavelength']) / 1000:.3f}'
```
I'm not even certain we require the `float()`... do we?

test_provider.py
sraaphorst: Here's the definitive reply from the Observing Tool: an organizational folder is just for human convenience and holds no relevance, so when you encounter an organizational folder, we could either just roll it up to the parent group and include the observations in it there as `AndGroup`s containing the `Observation`s, or create an `AndGroup` and create sub `AndGroup`s with the `Observation`s in them.
Screen Shot 2021-12-19 at 8 51 25 PM

test_provider.py
```diff
@@ -0,0 +1,310 @@
...
+        program = Program(json['programId'],
+                          json['key'],
+                          Band(int(json['queueBand'])),
+                          bool(json['isThesis']),
+                          ProgramMode(json['programMode']),
+                          None,
+                          None,
+                          ta,
+                          root_group,
+                          TooType(json['tooType']) if json['tooType'] != 'None' else None)
```
sraaphorst, line 299: I would really like us to stop relying on hard-coded string literals spread all over the code for things like this.

Let's try to keep them grouped to `Enum`s so that it's obvious where they are defined.

```
class ProgramLookups(Enum):
    ID = 'programId'
    KEY = 'key'
    BAND = 'queueBand'
....
json[ProgramLookups.ID]
....
```
Also, does the band come in as a string? Do we need to explicitly convert it to an `int`?

test_provider.py
```diff
@@ -0,0 +1,310 @@
...
+    
+    @staticmethod
+    def parse_time_allocation(json: dict) -> TimeAllocation:
+        #print(json['timeAccountAllocationCategories'])
+        ta = TimeAllocation(TimeAccountingCode(json['timeAccountAllocationCategories'][0]['category']),
+                            timedelta(milliseconds=json['awardedTime']),
+                            timedelta(milliseconds=0),
+                            timedelta(milliseconds=json['timeAccountAllocationCategories'][0]['programTime']),
+                            timedelta(milliseconds=json['timeAccountAllocationCategories'][0]['partnerTime']))
+        return ta
```
sraaphorst, line 246: More places where we should remove hardcoded `str`s.

Also, why are we just looking at index `[0]`? I'm not clear exactly what is happening here.

sraaphorst:
All in all, some good progress, with the following notes:

1. Please consult with me before making changes to the mini-model.
2. Let's remove hardcoded strings and make them into `Enum`s , possibly in nested classes (so we can keep them private, e.g:
```
class MyClass:
    class _ProgramKeys(Enum):
        ID = 'progId'
        ....
    ....
```
3. When instantiating `dataclass`es, read in the values first and then create the dataclass from its constructor as we might mark some of these `frozen` and we want them to usually be as immutable as possible: don't create the dataclass and then populate the values, since this will cause problems down the road, in all likelihood.
4. If you are having problems with any of the values, please continue to ask in the Scheduler Slack channel or Bryan or I... we will be happy to help however we can and the channel might be the best place, since Bryan can give the science view and I can give the programming view.
5. Let's avoid using `.format` strings and instead use `f`-strings to format.

sraaphorst (changes requested):
See notes.

odb_extractor_atoms.py
sraaphorst: Nix that last bit... it does look like we need the `float(...)` call as confirmed by the json file.

test_provider.py
```diff
@@ -0,0 +1,310 @@
...
+from common.api import ProgramProvider
+from common.minimodel import *
+
+
+class JsonProvider(ProgramProvider):
+
+    # TODO: To handle Enum it could be done this way or changing the Enum to accept the 
+    # string value and just use the constructor. Let me know which is better.
+
+    obs_classes = {'partnerCal': ObservationClass.PARTNER_CAL,
```
sraaphorst, line 12: For `ObservationClasses`, we do want the implicit ordering that `auto` gives us, so there are two ways I can see this going. The first way is this way, which is fine. The second way requires a bit more thought, so let me get back to you on that Sunday or Monday.

test_provider.py
```diff
@@ -0,0 +1,310 @@
+import json
+from datetime import datetime, timedelta
+from common.api import ProgramProvider
+from common.minimodel import *
+
+
+class JsonProvider(ProgramProvider):
```
sraaphorst, line 7: Why is this in a `test_provider.py` file? Isn't this the deliverable?

sraaphorst:
This looks good! Thank you for making the changes. I still think we need to add handling for organizational folders and some fields, and I'd like to get rid of the hard-coded strings, but good stuff.

PR #45, opened 2021-12-21
Sched 76: API implementation

stroncod:
Third time is the charm LOL

I add the Enums to access the JSON without using hard strings and the Organizational Folder support.


common/minimodel/__init__.py
```diff
@@ -437,7 +437,7 @@ class TooType(IntEnum):
     STANDARD = auto()
 
 
-class SetupTimeType(Enum):
+class SetupTimeType(IntEnum):
```
sraaphorst, line 440: My only concern with these is that I think we should use `Enum` if the ordering on them does not matter (i.e. we will not be sorting on these values or doing other ordering based comparisons) and `IntEnum` if the ordering is important: I was using this information to "tag" whether or not it was the case, but I think that this probably merits explicit documentation (on my part) in the actual class docstrings.

odb_extractor_atoms.py
```diff
@@ -160,7 +160,9 @@ def findatoms(observation):
             atoms[-1]['part_time'] += step_time
         else:
             atoms[-1]['prog_time'] += step_time
-
+            
+#         atoms[-1]['id'] = atomlabel
+            
         print('{:22} {:12} {:7.2f} {:3d} {:10} {:15} {:12} {:12} {:7.4f} {:5.2f} {:5.2f} {:3d}'.format(datalab,
```
sraaphorst, line 166: Again, just a nitpick, so we don't have to do this right away, but for the future, let's keep away from `.format` and use `f`-strings instead, as this is the best practice moving forward from what I can see.

test_provider.py
```diff
@@ -0,0 +1,454 @@
...
+
+    program_types = {'C': ProgramTypes.C,
+                     'CAL': ProgramTypes.CAL,
+                     'DD': ProgramTypes.DD,
+                     'DS': ProgramTypes.DS,
+                     'ENG': ProgramTypes.ENG,
+                     'FT': ProgramTypes.FT,
+                     'LP': ProgramTypes.LP,
+                     'Q': ProgramTypes.Q,
+                     'SV': ProgramTypes.SV}
```
sraaphorst, line 28: In this case, I don't think we need this `dict` since the correspondence is exact between the `ProgramTypes` and the codes. If we have a `str` value, say `program_type`, we can just do:
```
ProgramTypes[program_type]
```
and then avoid the `dict` completely.

test_provider.py
```diff
@@ -0,0 +1,454 @@
...
+    class _MagnitudeKeys(Enum):
+        NAME = 'name'
+        VALUE = 'value'
+
+    def __init__(self, path):
+        self.path = path
+
+    def load_program(self, path: str) -> dict:
+        with open(path, 'r') as f:
+            return json.loads(f.read())
```
sraaphorst, line 138: This seems like it should be a `@staticmethod`?

test_provider.py
```diff
@@ -0,0 +1,454 @@
...
+        value = json['value']
+        error = None
+        return Magnitude(band, value, error)
+    
+    @staticmethod
+    def get_program_dates(program_type: ProgramTypes, id: str, notes: list) -> Tuple[datetime, datetime]:
+        start = end = None
+        year = int(id[3:7])
+        next_year = year + 1
+        semester = id[7]
```
sraaphorst, line 160:
```
semester = SemesterHalf[id[7]]
```
Then we can work with `SemesterHalf.A` and `SemesterHalf.B` instead of `str` as we do up ahead.

Let's try to minimize our actual work with strings and push it as far back as it can go, so we are working with mini-model objects.

test_provider.py
```diff
@@ -0,0 +1,454 @@
...
+        if program_type is ProgramTypes.FT:
+
+            for note in notes:
+                if 'title' in note.keys():
+                    title = note['title'].lower()
+                    if 'cycle' in title or 'active' in title:
```
sraaphorst, line 166: Again, we should get `'title'` out of here.

test_provider.py
```diff
@@ -0,0 +1,454 @@
...
+                                months = field.split('-')
+                                if len(months) == 3:
+                                    break
+                        if len(months) == 0:
+                            for field in fields:
+                                f = field.lower().strip(' ,')
+                                if f in ['jan', 'feb', 'mar', 'apr', 'may', 'jun', 'jul', 'aug', 'sep', 'oct', 'nov', 'dec']:
+                                    months.append(f)
+                        im1 = months[0]
+                        im2 = months[-1]
```
sraaphorst, line 180: What does `im` mean?

test_provider.py
```diff
@@ -0,0 +1,454 @@
...
+                            start = datetime(year, im1, 1)
+                            end = datetime(year, im2, 1) if im2 > im1 else datetime(next_year, im2, 1)
+                        break
+        else:
+            if semester == 'A':
+                start = datetime(int(year), 2, 1)
+                end = datetime(int(year), 7, 31)
+            else:
+                start = datetime(int(year), 8, 1)
+                end = datetime(int(next_year), 1, 31)
```
sraaphorst, line 194: If there is some kind of error in the `semester` as specified, it will never be caught, which is why I think it's important to work with `SemesterHalf` and do the lookup instead of comparing `str`. If we attempt to do a lookup on a `SemesterHalf` that doesn't exist, then a `KeyError` will automatically be `raise`d. Right now, if an illegal `semester` is encountered, it will just always be treated as `B`.

test_provider.py
```diff
@@ -0,0 +1,454 @@
...
+        tw_arr = []
+        for timing_window in json[JsonProvider._TimingWindowsKeys.TIMING_WINDOWS.value]:
+            tw = TimingWindow()
+            repeat = JsonProvider._TimingWindowsKeys.REPEAT.value
+            tw.start = datetime.fromtimestamp(timing_window[JsonProvider._TimingWindowsKeys.START.value]/1000)
+            tw.duration = timedelta(milliseconds=timing_window[JsonProvider._TimingWindowsKeys.DURATION.value])
+            tw.repeat = timing_window[repeat]
+            tw.period = timedelta(milliseconds=timing_window[repeat]) if timing_window[repeat] > 0 else None
+            tw_arr.append(tw)
+        return tw_arr
```
sraaphorst, line 207: We want to retrieve the data and THEN create the `dataclass`:
```
repeat = ...
start = ...
duration = ...
...
tw = TimingWindow(...) # put fields in here
tw_arr.append(tw)
```

Dataclasses should be considered mutable and we should try to only populate them at the time they are constructed: we should not be constructing them and then populating them.

test_provider.py
```diff
@@ -0,0 +1,454 @@
...
+    @staticmethod
+    def parse_nonsidereal_target(json: dict) -> NonsiderealTarget:
+        base = JsonProvider._TargetKeys.BASE.value
+        return NonsiderealTarget(json[base][JsonProvider._TargetKeys.DES.value],
+                                 json[base][JsonProvider._TargetKeys.TAG.value],
+                                 None,
+                                 None)
+    
+    @staticmethod
+    def parse_atoms(sequence: List[dict], qa_states: List[QAState]) -> List[Atom]:
```
sraaphorst, line 246: I'm going to have to look at this in more detail and try an example.

test_provider.py
```diff
@@ -0,0 +1,454 @@
...
+        return or_group
+
+    @staticmethod
+    def parse_guide_star(json: dict) -> Resource:
+
+        # TODO: Maybe a GuideStart class should be in place to handle this better
+        res = Resource(json['name'], json['tag'])
+        return res
+    
+    def parse_root_group(self, json: dict) -> AndGroup:
```
sraaphorst, line 388: I think these methods should all be static, and calling each other statically.

test_provider.py
```diff
@@ -0,0 +1,454 @@
...
+                    groups.append(self.parse_or_group(json[key]))
+        num_to_observe = len(groups)
+        root_group = AndGroup(None, None, num_to_observe, 0, 0, groups, AndOption.ANYORDER)
+        return root_group
+
+    def parse_and_group(self, json: dict) -> AndGroup:
+        observations = [self.parse_observation(json[key], key) for key in json.keys() if key.startswith(self._ObsKeys.KEY.value)]
+
+        number_to_observe = len(observations)
+        delay_max, delay_min = 0, 0 # TODO: What are these?
```
sraaphorst, line 403: These have to do with cadences, I believe. Bryan goes into some detail on them in his document. I don't think we have to worry about them too much in the initial implementation, but if the information is available, we should populate it. They will likely become more important as we connect things to GPP.

test_provider.py
```diff
@@ -0,0 +1,454 @@
...
+                       json[self._ProgramKeys.INTERNAL_ID.value],
+                       Band(int(json[self._ProgramKeys.BAND.value])),
+                       bool(json[self._ProgramKeys.THESIS.value]),
+                       ProgramMode[json[self._ProgramKeys.MODE.value].upper()],
+                       program_type,
+                       start,
+                       end,
+                       ta,
+                       root_group,
+                       too_type)
```
sraaphorst, line 428: This is good, and how I'd like to see the mini-model classes instantiated and returned.

test_provider.py
```diff
@@ -0,0 +1,454 @@
...
+    
+    @staticmethod
+    def too_type_for_obs(program: Program) -> NoReturn:
+
+        for group in program.root_group.children:
+            for obs in group.children:
+                if program.too_type is TooType.STANDARD:
+                    obs.too_type = TooType.STANDARD
+                elif program.too_type is TooType.RAPID:
+                    obs.too_type = TooType.RAPID if json['tooOverrideRapid'] else TooType.STANDARD
```
sraaphorst, line 438: So the idea is to flow the `ToOType` down from the `Program` into the `Group`s unless overridden explicitly? This should probably have a recursive method inside of it to handle this that does not assume that we will limit the group structure. I'd like to take a look at this after the meeting today.

test_provider.py
sraaphorst: I had to change this around: I intended this method to parse out a single `TimingWindow` and not the entire list.

test_provider.py
sraaphorst: I ended up working on this method a bit, but I think the logic was well simplified over the XML: it just wasn't clear to me when I started, and I did add some error handling and divided it up a bit to make it more easily readable, but good job on cleaning this up.

PR #49, opened 2022-01-05
SCHED-76: API fixing and cleanup

sraaphorst:
Some fixes and cleanup to the `OcsProgramProvider`.
This still needs testing so future changes may be coming.


api/ocs/__init__.py
```diff
@@ -320,134 +405,280 @@ def parse_atoms(sequence: List[dict], qa_states: List[QAState]) -> List[Atom]:
...
-                              timedelta(milliseconds=json[OCSProgramProvider._TAKeys.CATEGORIES][0][OCSProgramProvider._TAKeys.PARTNER_TIME]))
+    def parse_target(data: dict) -> Target:
+        """
+        Parse a general target - either sidereal or nonsidereal - from the supplied data.
+        """
+        tag = data[OcsProgramProvider._TargetKeys.TAG]
+        if tag == 'sidereal':
+            return OcsProgramProvider.parse_sidereal_target(data)
+        elif tag == 'nonsidereal':
+            return OcsProgramProvider.parse_sidereal_target(data)
```
stroncod, line 418: This should be non-sidereal I imagine?

stroncod (changes requested):

api/ocs/__init__.py
sraaphorst: Ooops! You are right. Good catch. Thanks!

stroncod (approved):

sraaphorst:
We will probably have to make some more changes to this to accommodate Bryan's new atom code, and we need to check Dec is being properly converted.

PR #50, opened 2022-01-05
Sched 81: Horizons Client 

stroncod:


horizons/__init__.py
```diff
@@ -0,0 +1,281 @@
...
+        if hms is None:
+            return None
+
+        h = float(hms[0])
+        m = float(hms[1])
+        s = float(hms[2])
+        hours = h + m / 60. + s / 3600.
+        rad = hours * np.pi / 12.
+
+        return rad
```
sraaphorst, line 59: This is all good stuff that we should probably move out of here and put in a package in `common.helpers` or something similar.

horizons/__init__.py
```diff
@@ -0,0 +1,281 @@
...
+    API to interact with the Horizons service
+    """
+
+    # A not-complete list of solar system major body Horizons IDs
+    bodies = {'mercury': '199', 'venus': '299', 'mars': '499', 'jupiter': '599', 'saturn': '699',
+               'uranus': '799', 'neptune': '899', 'pluto': '999', 'io': '501'}
+
+    def __init__(self,
+                 site: Site,
+                 path: str = 'horizons/data',
```
sraaphorst, line 73: One comment: we should be building paths with `os.path.join('horizons', 'data')` as it's platform independent.

horizons/__init__.py
```diff
@@ -0,0 +1,281 @@
...
+    @staticmethod
+    def _angular_distance(ra1: float, dec1: float, ra2: float, dec2: float) -> float:
+        """
+        Calculate the angular distance between two points on the sky.
+        based on https://github.com/gemini-hlsw/lucuma-core/blob/master/modules/core/shared/src/main/scala/lucuma/core/math/Coordinates.scala#L52
+        """
+        φ1 = dec1
+        φ2 = dec2
+        delta_φ = dec2 - dec1
+        delta_λ = ra2 - ra1
```
sraaphorst, line 97: I did not know that Python supported variables starting with unicode characters. Interesting!

sraaphorst: Oh, question: why are you making references to `dec1` and `dec2` instead of using them directly?

horizons/__init__.py
```diff
@@ -0,0 +1,281 @@
...
+        self.site = site
+     
+    
+    @staticmethod
+    def generate_horizons_id(designation: str) -> str:
+        des = designation.lower()
+        return HorizonsClient.bodies[des] if des in HorizonsClient.bodies else designation
+    
+    @staticmethod
+    def _angular_distance(ra1: float, dec1: float, ra2: float, dec2: float) -> float:
```
sraaphorst, line 89: Since we're using `numpy`, it would probably be useful to make all of these types:
`Union[npt.NDArray[float], float]`
which will require the include:
`import numpy.typing as npt`
That way, we can use this function both with `float`s and with `numpy`  arrays of `float`.
This might be another one we want to consider moving out of here to `common.helpers` or similar.

horizons/__init__.py
```diff
@@ -0,0 +1,281 @@
...
+        # calculate angular distance to radians
+        delta = HorizonsClient.angular_distance(ra1, dec1, ra2, dec2)
+        if delta == 0:
+            return ra1, dec1
+        else:
+
+            φ1 = dec1
+            φ2 = dec2
+            λ1 = ra1
+            λ2 = ra2
```
sraaphorst, line 116: Again, I'd probably just sub these into the equations below instead of making references to them.

horizons/__init__.py
```diff
@@ -0,0 +1,281 @@
...
+        """
+        φ1 = dec1
+        φ2 = dec2
+        delta_φ = dec2 - dec1
+        delta_λ = ra2 - ra1
+        a = np.sin(delta_φ / 2)**2 + np.cos(φ1) * np.cos(φ2) * np.sin(delta_λ / 2)**2
+        return 2 * np.arctan2(np.sqrt(a), np.sqrt(1 - a))
+
+    @staticmethod
+    def interpolate(ra1: float, dec1: float, ra2: float, dec2: float, f: float) -> float:
```
sraaphorst, line 102: Also, this is another method we could make:
`Union[npt.NDArray[float], float]`
for all the types and `numpy` will handle both cases automatically for us.

horizons/__init__.py
```diff
@@ -0,0 +1,281 @@
...
+            name = f'DES={designation};'
+        else:
+            name = self.generate_horizons_id(designation)
+        return name
+
+    def _time_bounds(self) -> Tuple[str, str]:
+        """
+        Returns the start and end times based on the given date
+        """
+        return self.start.strftime('%Y%m%d_%H%M'), self.end.strftime('%Y%m%d_%H%M')
```
sraaphorst, line 142: We should probably extract `%Y%m%d_%H%M` as a static member of this class since we're using it twice.

horizons/__init__.py
```diff
@@ -0,0 +1,281 @@
...
+              airmass: int = 3) -> requests.Response:
+     
+        # The items and order follow the JPL/Horizons batch example:
+        # ftp://ssd.jpl.nasa.gov/pub/ssd/horizons_batch_example.long
+        # and
+        # ftp://ssd.jpl.nasa.gov/pub/ssd/horizons-batch-interface.txt
+        # Note that spaces should be converted to '%20'
+        
+        skip_day = 'NO' if daytime else 'YES'
+        center = '568@399' if self.site is Site.GN else 'I11@399'
```
sraaphorst, line 169: I'm not familiar with this... are these specific to the sites? If so, I can move them to the mini-model where they should probably be and put them in `SiteInformation`. What do they represent?

horizons/__init__.py
```diff
@@ -0,0 +1,281 @@
...
+        lastline = lines.index('$$EOE') - 1
+        time = np.array([])
+        ra = np.array([])
+        dec = np.array([])
+
+        for line in lines[firstline:lastline]:
+            if line and line[7:15] != 'Daylight' and line[7:14] != 'Airmass':
+
+                values = line.split(' ')
+                values = d.split(' ')
```
sraaphorst, line 245: Not too sure what's going on here: the second command cancels out the first, since you're overwriting `values`?

sraaphorst: Is `d` even defined at this point? I don't see it anywhere.

sraaphorst:
So far, this looks good to me. I made some suggestions and asked for some clarification and investigation, but good stuff so far. 👍

horizons/__init__.py
stroncod: Those are specific to site: https://ssd.jpl.nasa.gov/horizons/manual.html#center

horizons/__init__.py
stroncod: So in the original code it require two transformation and I was not sure if we were going to need the same thing. Apparently is in the correct units (radians) per the transformation functions (now in helpers).

horizons/__init__.py
sraaphorst: I made a small PR to move these to the mini-model so that we can keep as much as the site information together so that the code is more easily adaptable to different observatories.

Once that's merged, you should just be able to use:
`self.site.coordinate_center`

Or reference it easier using:
```
center = self.site.coordinate_center
```
and then get rid of the `if` `else` on this line.

horizons/__init__.py
```diff
@@ -0,0 +1,260 @@
...
+import requests
+from typing import Tuple, Union
+import logging
+import contextlib
+import os
+
+@dataclass
+class EphemerisCoordinates:
+    """
+    Both ra and dec are in radians.
```
sraaphorst, line 17: It looks like we're going to be working with degrees, I believe, so we will have to do a small conversion here.

sraaphorst:
I don't see any issues here at a glance except for the fact that we will be working with decimal degrees and not radians in the Scheduler, so I think if you make that change and check to test, we can merge it.

We probably should be bringing some property-based testing into the Scheduler and incorporate the Hypothesis library.
It is very easy to use and will allow us to follow some GPP conventions and test this code in a way that is similar to what Shane posted to the #scheduler Slack group today. I'll make a Jira task for this.

horizons/__init__.py
```diff
@@ -0,0 +1,260 @@
...
+    def generate_horizons_id(designation: str) -> str:
+        des = designation.lower()
+        return HorizonsClient.bodies[des] if des in HorizonsClient.bodies else designation
+
+    @staticmethod
+    def interpolate(ra1: Union[npt.NDArray[float], float],
+                    dec1: Union[npt.NDArray[float], float],
+                    ra2: Union[npt.NDArray[float], float],
+                    dec2: Union[npt.NDArray[float], float],
+                    f: Union[npt.NDArray[float], float]) -> float:
```
sraaphorst, line 59: Awesome. Thanks for doing this!

common/helpers/helpers.py
```diff
@@ -24,3 +24,56 @@ def str_to_bool(s: Optional[str]) -> bool:
...
+
+    h = float(hms[0])
+    m = float(hms[1])
+    s = float(hms[2])
+    hours = h + m / 60. + s / 3600.
+    rad = hours * np.pi / 12.
+
+    return rad
+
+def angular_distance(ra1: float, dec1: float, ra2: float, dec2: float) -> float:
```
sraaphorst, line 68: Sorry to bug you about this, but could we also make this one use `Union[npt.NDArray[float], float]` since numpy will handle it automatically?

common/helpers/helpers.py
```diff
@@ -24,3 +24,56 @@ def str_to_bool(s: Optional[str]) -> bool:
...
+    d = float(dms[0])
+    m = float(dms[1])
+    s = float(dms[2])
+    sign = dms[3]
+    dd = d + m / 60. + s / 3600.
+
+    if sign == '-':
+        dd *= -1.
+
+    return dd
```
sraaphorst, line 44: I guess we should probably have typing information here?
`dms: Optional[List[float]]`
with a return type of `float`.

common/helpers/helpers.py
```diff
@@ -24,3 +24,56 @@ def str_to_bool(s: Optional[str]) -> bool:
...
+
+
+def dms2rad(dms):
```
sraaphorst, line 47: Same here re: typing.

We might not even need these since we're going to be working to be working with decimal degrees, so it's possible we can eliminate them instead of converting into radians and then converting back, which is what we will have to do now, but maybe the equations for interpolation require radians explicitly, in which case, best not try to mess with them.

common/helpers/helpers.py
```diff
@@ -24,3 +24,56 @@ def str_to_bool(s: Optional[str]) -> bool:
...
+    if dms is None:
+        return None
+    
+    dd = dms2deg(dms)
+    rad = dd * np.pi / 180.
+    return rad
+
+
+def hms2rad(hms):
```
sraaphorst, line 56: Same as above re: typing and decimal degrees.

PR #51, opened 2022-01-06
SCHED-80: Add coordinate centers to SiteInformation

sraaphorst:
In order to limit the site-specific information to the mini-model so that it is easier to adapt this code to other observatories, we move the coordinate centers to the `SiteInformation` class:

https://ssd.jpl.nasa.gov/horizons/manual.html#center


stroncod:
It looks like there are repeated commits from PR #49 ?? I assume it won't be an issue.

PR #57, opened 2022-01-21
Sched 87: Property-based Test for Horizons

stroncod:
3 of the 6 test failed as I think there is a problem with the signed microarc seconds transformation that I wan to discuss with Sebastian (or been reviewed here) that makes those test fail. 
There is also a unit test added to test the responde from the client itself (either by query or by file). 

This branch has SCHED-81 as base so this probably would bring those changes too if merged.


horizons/__init__.py
```diff
@@ -0,0 +1,307 @@
...
+ 
+        horizons_name = self._form_horizons_name(target.tag, target.des)
+        logging.info(f'{target.des}')
+
+        
+        file = self._get_ephemeris_file(target.des) if target.tag is not TargetTag.MAJOR_BODY else self._get_ephemeris_file(horizons_name) 
+        
+        if not overwrite and os.path.exists(file):
+            logging.info(f'Saving ephemerides file for {target.des}')
+            print('AAAAAAAAA')
```
sraaphorst, line 260: 😄 I'm getting we can probably get rid of this line.

test_horizons_client.py
```diff
@@ -0,0 +1,125 @@
...
+        max_size=4,
+    )
+)
+def test_angular_distace_symmetry(values):
+    """
+    Angular Distance must be symmetric to within 1µas
+    """
+    a, b, c, d = values
+    delta = Coordinates(a, b).angular_distance(Coordinates(c, d)) - Coordinates(c, d).angular_distance(Coordinates(a, b))
+    assert Angle.to_signed_microarcseconds(delta) <= 1
```
sraaphorst, line 61: I was thinking about the signed microarcseconds... I need to look a bit more closely, but I think that these tests will fail because the data we get from OCS is not as precise as the data we get from GPP, so converting TO signed microarcseconds from a float will not offer the same precision. We should probably be doing tests that are less precise in these cases? When you are back from vacation, let's discuss what is failing and why: either that, or I will look at it this week.

tests/test_horizons_client.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # "interpolate should result in angular distance of 0° from `b` for factor 1.0, within 1µsec (15µas)"
+    delta = helpers.angular_distance(b, HorizonsClient.interpolate(a, b, 1.0))
+    assert abs(toMicroArcSeconds(delta)) < 15
+
+@given(a=st.floats, b=st.floats)
+def test_interpolation_by_fractional_angular_separation(a, b):
+
+    µas180 = ...
+    µas360 = ...
```
sraaphorst, line 27: I'm not sure what is supposed to be going on here?

tests/test_horizons_client.py
```diff
@@ -0,0 +1,39 @@
...
+def test_interpolation_by_fractional_angular_separation(a, b):
+
+    µas180 = ...
+    µas360 = ... 
+
+    sep = angular_distance(a, b)
+    deltas = [ ]
+
+    for f in range(-1,2):
+        (f/10.0)
```
sraaphorst, line 33: ...or here. 🤔

horizons/__init__.py
```diff
@@ -0,0 +1,307 @@
...
+        Convert an angle in radians to a signed degree angle.
+        """
+        return angle * 180.0 / np.pi
+
+    @staticmethod
+    def to_microarcseconds(angle: float) -> float:
+        """
+        Convert an angle in radians to a signed microarcsecond angle.
+        """
+        return Angle.to_degrees(angle) * Angle.µasPerDegree
```
sraaphorst, line 43: Yes... this is probably what is causing the issues, I would think... major loss of precision here over GPP's representation of angles, which would cause our test cases to fail since we're using `float` and they are using `Long` and storing everything directly (I believe) in microarcseconds whereas we are storing in signed decimal degrees, which is much less precise.

horizons/__init__.py
```diff
@@ -17,7 +17,7 @@ class Angle:
     """
     Angle in radians.
     """
-    µasPerDegree: float = 60 * 60 * 1000 * 1000
+    µasPerDegree: float =  60 * 1000
```
sraaphorst, line 20: The name and conversion here definitely isn't right. We have that there are indeed:
```
60 * 60 * 1000 * 1000
```
microarcseconds per degree as per the original line.

By eliminating a factor of 60 (?) and a factor of 1000, I am confused what unit we get here.

If we had:
```
60 * 60 * 1000
```
then we would have milliarcseconds per degree.

By removing a factor of 60, though, I'm not sure what this is supposed to be: something in arcminutes instead of arcseconds. I think maybe "milliarcminutes," which isn't really a unit that is used?

We should maybe make this:
```
milliarcsecondsPerDegree = 60 * 60 * 1000
```
and then change all references from microarcseconds to milliarcseconds?

sraaphorst: Of course, this will require changes to propagate to the test cases... references to microarcseconds should be converted to milliarcseconds, and you may need to recalculate some of the values.

Is everything passing now? I'm a bit confused by what's going on.

sraaphorst (approved):
Looks great!

PR #62, opened 2022-02-02
Refactored, latest atom algorithm

bryanmiller:
Refactored and updated atom algorithm handling NIR imaging and do not split.
Calculates the observation mode (imaging, longslit, mos, ifu, etc). 
Option for XLSX output and manual atom edits.

### DIFF
--- a/odb_extractor_atoms.py
+++ b/odb_extractor_atoms.py
@@ -6,15 +6,179 @@
 # 2021-11-24
 
 import os
+import sys
 import json
 import gzip
 import numpy as np
+import matplotlib.pyplot as plt
+from scipy.signal import find_peaks
+import requests
 
 from openpyxl import Workbook
 from openpyxl import load_workbook
 
 
-fpuinst = {'GNIRS': 'instrument:slitWidth', 'GMOS-N': 'instrument:fpu'}
+fpuinst = {'GSAOI': 'instrument:utilityWheel', 'GPI': 'instrument:observingMode', 'Flamingos2': 'instrument:fpu',
+           'NIFS': 'instrument:mask', 'GNIRS': 'instrument:slitWidth', 'GMOS-N': 'instrument:fpu',
+           'GMOS-S': 'instrument:fpu', 'NIRI': 'instrument:mask'}
+
+gpi_filter_wav = {'Y': 1.05, 'J': 1.25, 'H': 1.65, 'K1': 2.05, 'K2': 2.25}
+nifs_filter_wav = {'ZJ': 1.05, 'JH': 1.25, 'HK': 2.20}
+
+
+def find_filter(input, filter_dict):
+    """Match input string with filter list (in dictionary)"""
+
+    filter = ''
+    filters = list(filter_dict.keys())
+    for filt in filters:
+        if filt in input:
+            filter = filt
+            break
+    return filter
+
+
+def uniquelist(seq):
+    """Make a list of unique values"""
+    # http://stackoverflow.com/questions/480214/how-do-you-remove-duplicates-from-a-list-in-python-whilst-preserving-order
+    seen = set()
+    seen_add = seen.add
+    return [x for x in seq if not (x in seen or seen_add(x))]
+
+
+def searchlist(val, alist):
+    """Search for existence of val in any element of alist"""
+    found = False
+    for elem in alist:
+        if val in elem:
+            found = True
+            break
+    return found
+
+
+def shortid(idin):
+    """Return short form of obsid or data label"""
+    idsp = idin.split('-')
+    #         print(obsidsp)
+    idout = idsp[0][1] + idsp[1][2:5] + '-' + idsp[2] + '-' + idsp[3] + '[' + idsp[4] + ']'
+    if len(idsp) == 6:
+        idout += '-' + idsp[5]
+    return idout
+
+
+def odb_json(progid, path='None', overwrite=False, verbose=False):
+    """
+    Download json of ODB program information
+
+    Parameters
+        progid:  Program ID of program to extract
+        path:    Path for json files
+        overwrite: Overwrite any existing json files?
+        verbose: Verbose output?
+
+    Return
+        json_result:   JSON query result as a list of dictionaries
+    """
+
+    if progid == "":
+        print('odb_json: program id not given.')
+        raise ValueError('Program id not given.')
+
+    file = progid + '.json.gz'
+    if not overwrite and path != 'None' and os.path.exists(os.path.join(path, file)):
+        with gzip.open(os.path.join(path, file), 'r') as fin:
+            json_bytes = fin.read()
+
+        json_str = json_bytes.decode('utf-8')
+        json_result = json.loads(json_str)
+    else:
+        response = requests.get(
+            'http://gnodbscheduler.hi.gemini.edu:8442/programexport?id=' + progid)
+        try:
+            response.raise_for_status()
+        except requests.exceptions.HTTPError as exc:
+            print('odb_json: request failed: {}'.format(response.text))
+            raise exc
+        else:
+            json_result = response.json()
+            if (overwrite or not os.path.exists(os.path.join(path, file))) and path != 'None':
+                json_str = json.dumps(json_result, indent=2)
+                with gzip.open(os.path.join(path, file), 'wb') as fout:
+                    fout.write(json_str.encode('utf-8'))
+
+        if verbose:
+            print(response.url)
+        # print(response.text)
+
+    return json_result
+
+# --------------
+
+
+def obsmode(config):
+    """Determine the observation mode (e.g. imaging, longslit, mos, ifu,..."""
+
+    mode = 'unknown'
+    if searchlist('GMOS', config['inst']):
+        if 'MIRROR' in config['disperser']:
+            mode = 'imaging'
+        elif searchlist('arcsec', config['fpu']):
+            mode = 'longslit'
+        elif searchlist('IFU', config['fpu']):
+            mode = 'ifu'
+        elif 'CUSTOM_MASK' in config['fpu']:
+            mode = 'mos'
+    elif config['inst'] in ["GSAOI", "'Alopeke", "Zorro"]:
+        mode = 'imaging'
+    elif config['inst'] in ['IGRINS', 'MAROON-X']:
+        mode = 'longslit'
+    elif config['inst'] in ['GHOST', 'MAROON-X', 'GRACES', 'Phoenix']:
+        mode = 'xd'
+    elif config['inst'] == 'Flamingos2':
+        if searchlist('LONGSLIT', config['fpu']):
+            mode = 'longslit'
+        if searchlist('FPU_NONE', config['fpu']) \
+                and searchlist('IMAGING', config['disperser']):
+            mode = 'imaging'
+    elif config['inst'] == 'NIRI':
+        if searchlist('NONE', config['disperser']) and searchlist('MASK_IMAGING', config['fpu']):
+            mode = 'imaging'
+    elif config['inst'] == 'NIFS':
+        mode = 'ifu'
+    elif config['inst'] == 'GNIRS':
+        if searchlist('mirror', config['disperser']):
+            mode = 'imaging'
+        elif searchlist('XD', config['disperser']):
+            mode = 'xd'
+        else:
+            mode = 'longslit'
+    elif config['inst'] == 'GPI':
+        if searchlist('CORON', config['fpu']):
+            mode = 'coron'
+        elif searchlist('NRM', config['fpu']):
+            mode = 'nrm'
+        elif searchlist('DIRECT', config['fpu']):
+            mode = 'imaging'
+        else:
+            mode = 'ifu'
+
+    return mode
+
+# --------------
+
+
+def guide_state(step):
+    """Determine if guiding is on/off for a sequence step"""
+    # One could also extract the guider used if needed
+    guiding = False
+    for key in list(step.keys()):
+        if 'guideWith' in key:
+            if step[key] == 'guide':
+                guiding = True
+                break
+    return guiding
+
+# --------------
 
 
 def select_qastate(states):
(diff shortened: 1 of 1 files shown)


odb_extractor_atoms.py
```diff
@@ -6,15 +6,179 @@
...
+def find_filter(input, filter_dict):
+    """Match input string with filter list (in dictionary)"""
+
+    filter = ''
+    filters = list(filter_dict.keys())
+    for filt in filters:
+        if filt in input:
+            filter = filt
+            break
+    return filter
```
sraaphorst, line 38: These are things that @stroncod or I can change in the implementation if necessary, but this implementation would be more efficient, and I would think we would want to return `None` instead of the empty string if nothing was found? (If not, we can change the `None` to the empty string in the second line and remove the `Optional` type hint.)

```
def find_filter(input: str, filter_dict: Mapping[str, str]) -> Optional[str]:
    return next(filter(lambda f: f in input, filter_dict), None)
```

odb_extractor_atoms.py
```diff
@@ -6,15 +6,179 @@
...
+
+
+def searchlist(val, alist):
+    """Search for existence of val in any element of alist"""
+    found = False
+    for elem in alist:
+        if val in elem:
+            found = True
+            break
+    return found
```
sraaphorst, line 56: I think this does the trick and avoids the for loop.
```
def searchlist(val, alist):
    return any(val in elem for elem in alist)
```

odb_extractor_atoms.py
```diff
@@ -6,15 +6,179 @@
...
+        if verbose:
+            print(response.url)
+        # print(response.text)
+
+    return json_result
+
+# --------------
+
+
+def obsmode(config):
```
sraaphorst, line 118: I would think we would want `mode` to be an `Enum` and not a `str`?

```
class ObservationMode(str, Enum):
    UNKNOWN = 'unknown'
    IMAGING = 'imaging'
    LONGSLIT = 'longslit'
    IFU = 'ifu'
    MOS = 'mos'
    XD = 'xd'
    CORON = 'coron'
    NRM = 'nrm'

    @staticmethod
    def obsmode(config):
        # Put all obsmode logic in here
        ...
```

Then we have all of this self-contained, and can just call:
```
ObservationMode.obsmode(config)
```
which would be nice and clean.

sraaphorst: Also, do we want `unknown` or could we use `None` instead?

odb_extractor_atoms.py
```diff
@@ -6,15 +6,179 @@
...
+def guide_state(step):
+    """Determine if guiding is on/off for a sequence step"""
+    # One could also extract the guider used if needed
+    guiding = False
+    for key in list(step.keys()):
+        if 'guideWith' in key:
+            if step[key] == 'guide':
+                guiding = True
+                break
+    return guiding
```
sraaphorst, line 179:
```
def guide_state(step: dict) -> bool:
    return any('guideWith' in key and value == 'guide' for key, guide in step.items())
```

odb_extractor_atoms.py
```diff
@@ -59,98 +223,297 @@ def select_obsclass(classes):
...
+    do_not_split = False
+    exptimes = []
+    coadds = []
+    qoffsets = []
+    poffsets = []
+    qoffsets_sky = []
+    poffsets_sky = []
+    config = {'inst': '', 'fpu': [], 'disperser': [], 'filter': [], 'wavelength': []}
+    for ii, step in enumerate(sequence):
+        step_keys = list(step.keys())
```
sraaphorst, line 284: Just a Python coding tip: we should not have to do this, and taking a `dict_keys` object and converting it to a  `list` is (probably) quite inefficient: I am not completely certain of how Python handles this, but I imagine that it is an `O(n)` operation both in execution and memory use where `n` is the number of entries in `step`.

We can just ask:
```
if 'instrument:disperser' in step:
     ...
```
which automatically looks at the keys only to see if they contain `'instrument_disperser'` instead of doing:
```
step_keys = list(step.keys())
...
if 'instrument:disperser' in step_keys:
    ....
```

bryanmiller: Thanks, I'll try to remember that

odb_extractor_atoms.py
```diff
@@ -59,98 +223,297 @@ def select_obsclass(classes):
...
+        elif inst in ['IGRINS', 'MAROON-X']:
+            disperser = inst
+        else:
+            disperser = 'None'
+        if inst == 'GNIRS':
+            if step['instrument:acquisitionMirror'] == 'in' and \
+                    step['instrument:decker'] == 'acquisition':
+                disperser = 'mirror'
+            else:
+                disperser = disperser.strip('grating') + step['instrument:crossDispersed']
```
sraaphorst, line 310: Is the `disperser.strip('grating')` really what we want? Assuming `disperser` is a `str`, then this just removes any leading or trailing characters in:
```
['g', 'r', a', t', 'i', 'n']
```
from `disperser` and returns the result, which doesn't intuitively seem like what we would want to do.

Including the last `g` in `grating` isn't necessary since it treats this as a set of characters to remove from the front and end of the value contained in `disperser`.

Looking at the output for:
http://gnodbscheduler:8442/programexport?id=GN-2018B-Q-101

and seeing the entries in the sequence such as:
```
"instrument:disperser" : "32 l/mm grating"
```
I can see why this implementation does (I suspect) what we want it to do, but I think what we actually want is:
```
disperser.replace('grating', '')
```

Then, if we have:
```
disperser = "32 l/mm grating"
step['instrument:crossDispersed'] = 'SXD'
```
Doing:
```
disperser = disperser.replace('grating', '') + step['instrument:crossDispersed']
```
gives us:
```
disperser = '32 l/mm SXD'
```
which is - I assume - what we want?

odb_extractor_atoms.py
```diff
@@ -59,98 +223,297 @@ def select_obsclass(classes):
...
 
+    # First pass to analyze offsets and exptimes/coadds
+    do_not_split = False
+    exptimes = []
+    coadds = []
+    qoffsets = []
+    poffsets = []
+    qoffsets_sky = []
+    poffsets_sky = []
+    config = {'inst': '', 'fpu': [], 'disperser': [], 'filter': [], 'wavelength': []}
```
sraaphorst, line 282: If this is the `config` structure that we are expecting, i.e. we know the fields that are going to be in `config`, I suggest that we make this an actual class instead:
```
@dataclass
class InstrumentConfig:
    inst: List[str]
    fpu: List[str]
    disperser: List[str]
    filter: List[str]
    wavelength: List[str] # perhaps this should be float?
```
Then declare here:
```
inst = []
fpu = []
disperser = []
filter = []
wavelength = []
```
and after we have gathered the information into these, create the `InstrumentConfig` with:
```
InstrumentConfig(inst=inst, fpu=fpu, disperser=disperser, filter=filter, wavelength=wavelength)
```
(We could obviously just omit the keywords there, but it's generally recommended to include them in dataclasses.)

odb_extractor_atoms.py
```diff
@@ -59,98 +223,297 @@ def select_obsclass(classes):
...
+        else:
+            if inst == 'GNIRS':
+                filter = 'None'
+            else:
+                filter = 'Unknown'
+        if inst == 'NIFS' and 'Same as Disperser' in filter:
+            for filt in list(nifs_filter_wav.keys()):
+                if disperser[0] in filt:
+                    filter = filt
+                    break
```
sraaphorst, line 329:
```
filter = next(filter(lambda filt: disperser[0] in filt, nifs_filter_wav), None)
```

odb_extractor_atoms.py
```diff
@@ -59,98 +223,297 @@ def select_obsclass(classes):
...
+        config['filter'].append(filter)
```
sraaphorst, line 330: In general, this function is making me a bit nervous. I'm sure we can assume that all of these fields will exist in the sequence and be populated by the ODB, but I feel like the whole thing should be wrapped in a `try`...`except` block:
```
try:
   ...
except KeyError as e:
   ...
```
As it currently stands, if for some obscure reason, one of them doesn't exist, we are either going to get a corrupt data structure that will have unpredictable side effects, or we will get an uncaught error that is going to bubble up and crash the entire Scheduler.

odb_extractor_atoms.py
```diff
@@ -59,98 +223,297 @@ def select_obsclass(classes):
...
+        config['wavelength'].append(wavelength)
+
+        p = 0.0
+        q = 0.0
+        # Just want exposures on sky for dither pattern analysis
+        if step['observe:observeType'].upper() not in ['FLAT', 'ARC', 'DARK', 'BIAS']:
+            if 'telescope:p' in step_keys:
+                p = float(step['telescope:p'])
+            else:
+                p = 0.0
```
sraaphorst, line 345: Repeated strings like `telescope:p` should probably be defined somewhere since "magic strings" like this are considered to be a bad idea.

While it isn't strictly necessary, this could all be simplified to:
```
p = float(step['telescope:p']) if 'telescope:p' in step else 0.0
```

odb_extractor_atoms.py
```diff
@@ -59,98 +223,297 @@ def select_obsclass(classes):
...
+        for log_entry in observation['obsLog']:
+            obslog[log_entry['label']] = {'qaState': log_entry['qaState'], 'filename': log_entry['filename']}
+        datalabels = list(obslog.keys())
 
     # Sequence analysis
     sequence = observation['sequence']
     nsteps = len(sequence)
 
+    # First pass to analyze offsets and exptimes/coadds
+    do_not_split = False
```
sraaphorst, line 275: We don't need this here: we should move it down to where it is used. See my suggestion below.

odb_extractor_atoms.py
```diff
@@ -59,98 +223,297 @@ def select_obsclass(classes):
...
+        exptimes.append(float(step['observe:exposureTime']))
+        coadds.append(ncoadds)
+
+    mode = obsmode(config)
+    if verbose:
+        print(config)
+        print(mode)
+
+    if config['inst'] == 'GPI':
+        do_not_split = True
```
sraaphorst, line 370: Recommend we eliminate this...

odb_extractor_atoms.py
```diff
@@ -59,98 +223,297 @@ def select_obsclass(classes):
...
+
+    # Analyze sky offset patterns using auto-correlation
+    # The lag is the length of any pattern, 0 means no repeating pattern
+    plag = 0
+    qlag = 0
+    offset_lag = 0
+    if do_not_split:
```
sraaphorst, line 377: ...and then just change this to:
```
if config['inst'] == 'GPI':
```
I don't see other uses of the `do_not_split` variable so unless some will be added, it seems unnecessary, and this contains the logic to a smaller area, making it more comprehensible.

odb_extractor_atoms.py
```diff
@@ -59,98 +223,297 @@ def select_obsclass(classes):
...
+
+    # Second pass to determine atom properties
+    # Write results to the Excel worksheet if given
+    if ws is not None:
+        # Columns
+        columns = ['datalab', 'class', 'type', 'inst', 'exec_time', 'exptime', 'coadds', 'fpu', 'filter',
+                   'disperser', 'wavelength', 'p', 'q', 'guiding', 'qa_state', 'atom']
+        row = 1
+        for jj, col in enumerate(columns):
+            _ = ws.cell(column=jj + 1, row=row, value="{0}".format(col))
```
sraaphorst, line 419: Why assign at all if we're not interested in the return value? Couldn't we just write:
```
ws.cell(column=jj + 1, row=row, value="{0}".format(col))
```

sraaphorst: (Same comment applies for other uses of `ws.cell` below.)

odb_extractor_atoms.py
```diff
@@ -59,98 +223,297 @@ def select_obsclass(classes):
...
+    noffsets = 0
     for ii, step in enumerate(sequence):
         nextatom = False
-        qoffsets = False
 
         datalab = step['observe:dataLabel']
         if datalab in datalabels:
             qastate = obslog[datalab]['qaState']
         else:
             qastate = 'NONE'
```
sraaphorst, line 431: Through here and below, we have a `QAState` enum in the mini-model: I think we should be using these values?

odb_extractor_atoms.py
```diff
@@ -59,98 +223,297 @@ def select_obsclass(classes):
...
-        qastates.append(qastate.upper())
 
         observe_class = step['observe:class']
```
sraaphorst, line 433: Similar to `QAState`, am I correct in assuming this corresponds to the `ObservationClass` enum in the mini-model?

odb_extractor_atoms.py
```diff
@@ -59,98 +223,297 @@ def select_obsclass(classes):
...
                 nextatom = True
-                print('Atom for ABBA')
-        else:
-            nabba -= 1
-
+                atomstr += 'exposure time change, '
+
+            # Offsets - a new offset pattern is a new atom
+            #         print('npattern: ', npattern)
+            if not (offset_lag == 0 and exptime_groups == True):
```
sraaphorst, line 452: `bool` should never be explicitly checked, so it should just be:
```
if not(offset_lag == 0 and exptime_groups):
```

But even easier, we can distribute the `not` into the condition using de Morgan's law and just get:
```
if offset_lag or not exptime_groups:
```

odb_extractor_atoms.py
```diff
@@ -59,98 +223,297 @@ def select_obsclass(classes):
...
+                # For NIR imaging, need to have at least two offset positions if no repeating pattern
+                # New atom after every 2nd offset (noffsets is odd)
+                if mode == 'imaging' and offset_lag == 0 and all([w > 1.0 for w in config['wavelength']]):
```
sraaphorst, line 455: Another efficiency improvement: use a generator instead of a list in `any` and `all` expressions.
Instead of:
```
all([w > 1.0 for w in config['wavelength']])
```
just do:
```
all(w > 1.0 for w in config['wavelength'])
```
This avoids the overhead of allocating the memory to create a list and then populating it.

odb_extractor_atoms.py
```diff
@@ -59,98 +223,297 @@ def select_obsclass(classes):
...
+                    if ii == 0:
+                        noffsets += 1
+                    else:
+                        if poffsets[ii] != poffsets[ii_prevobj] or qoffsets[ii] != qoffsets[ii_prevobj]:
+                            noffsets += 1
+                    if noffsets % 2 == 1:
```
sraaphorst, line 461: It's not explicitly clear to me why `noffsets` being odd has the special case of starting a next atom?

odb_extractor_atoms.py
```diff
@@ -59,98 +223,297 @@ def select_obsclass(classes):
...
+                if verbose:
+                    print('QA states: ', qastates, file=fid)
+                    print('Classes: ', classes, file=fid)
+                    print('Guiding: ', guiding, file=fid)
+                atoms[-1]['guide_state'] = any(guiding)
+                atoms[-1]['wavelength'] = config['wavelength'][ii]
+                atoms[-1]['required_resources']['inst'] = config['inst']
+                atoms[-1]['required_resources']['filter'] = config['filter'][ii]
+                atoms[-1]['required_resources']['disperser'] = config['disperser'][ii]
+                atoms[-1]['required_resources']['fpu'] = config['fpu'][ii]
```
sraaphorst, line 490: This I find particularly confusing: why are we modifying the last atom in the list instead of creating a new one and appending it? I think that may be equivalent to what is happening with the `atoms.append` below, but it seems this could be done in a way that is clearer?

odb_extractor_atoms.py
```diff
@@ -59,98 +223,297 @@ def select_obsclass(classes):
...
             atoms.append({'id': natom, 'exec_time': 0.0, 'prog_time': 0.0, 'part_time': 0.0,
-                          'class': 'NONE', 'qastate': 'NONE'})
+                          'class': 'NONE', 'observed': False, 'qa_state': 'NONE', 'guide_state': False,
+                          'wavelength': 0.0,
+                          'required_resources': {'inst': 'NONE', 'filter': 'NONE', 'disperser': 'NONE',
+                                                 'fpu': 'NONE'}})
             classes = []
+            guiding = []
+            qastates = []
+            if step['observe:observeType'].upper() in ['FLAT', 'ARC', 'DARK', 'BIAS'] \
```
sraaphorst, line 508: The list should probably be defined elsewhere, because as it stands, we recreate it every time we reach this statement, which is unnecessary.

odb_extractor_atoms.py
```diff
@@ -59,98 +223,297 @@ def select_obsclass(classes):
...
+
+                # Print basic atom info
+                print(" \t exec_time: {:7.2f}, prog_time: {:7.2f}, part_time: {:7.2f}, guide_state: {}".
+                      format(atoms[-1]['exec_time'], atoms[-1]['prog_time'], atoms[-1]['part_time'],
+                             atoms[-1]['guide_state']), file=fid)
+
+            # New atom entry
+            print(atomstr, file=fid)
             natom += 1
             atoms.append({'id': natom, 'exec_time': 0.0, 'prog_time': 0.0, 'part_time': 0.0,
```
sraaphorst, line 500: Will we be using `Atom` from the mini-model here instead of a general `dict`?

odb_extractor_atoms.py
```diff
@@ -160,103 +523,173 @@ def findatoms(observation):
...
+        if ws is not None:
+            # Columns
+            #             columns = ['datalab', 'class', 'type', inst', 'exec_time', 'exptime', 'coadds', 'fpu', 'filter',
+            #                        'disperser', 'wavelength', 'p', 'q', 'guiding', 'qa_state', 'atom']
+            data = [datalab, observe_class.upper(), step['observe:observeType'].upper(), config['inst'],
+                    step_time, exptimes[ii], coadds[ii], config['fpu'][ii],
+                    config['filter'][ii], config['disperser'][ii], config['wavelength'][ii], poffsets[ii],
+                    qoffsets[ii], guiding[-1], qastate.upper(), atomlabel]
+            for jj in range(len(columns)):
+                _ = ws.cell(column=jj + 1, row=row, value=data[jj])
```
sraaphorst, line 542: `range(len(...))` is a Python anti-pattern. Use `enumerate` instead.

```
for jj, value in enumerate(columns):
    ws.cell(column=jj + 1, row=row, value=value)
```

odb_extractor_atoms.py
```diff
@@ -160,103 +523,173 @@ def findatoms(observation):
...
+        atoms[-1]['wavelength'] = config['wavelength'][ii]
+        atoms[-1]['required_resources']['inst'] = config['inst']
+        atoms[-1]['required_resources']['filter'] = config['filter'][ii]
+        atoms[-1]['required_resources']['disperser'] = config['disperser'][ii]
+        atoms[-1]['required_resources']['fpu'] = config['fpu'][ii]
+
+        # Print basic atom info
+        print(" \t exec_time: {:7.2f}, prog_time: {:7.2f}, part_time: {:7.2f}, guide_state: {}".
+              format(atoms[-1]['exec_time'], atoms[-1]['prog_time'], atoms[-1]['part_time'],
+                     atoms[-1]['guide_state']), file=fid)
```
sraaphorst, line 566: Again, it seems to me that here, we should be creating a new atom, appending it, and then outputting it.

Indeed, if we use the mini-model, we could just add a `def __str__(self)` method to `Atom` and format it exactly like this, and reduce the print to:
```
print(atom, file=fid)
```
Another note: in later versions of Python 3.9, use of `.format` is discouraged. Instead, we should be using `f`-strings:
```
print(f" \t exec_time: {atoms[-1]['exec_time']:7.2f}, prog_time: {atoms[-1]['prog_time']:7.2f}, part_time: {atoms[-1]['part_time']:7.2f}, guide_state: {atoms[-1]['guide_state']}", file=fid)
```

odb_extractor_atoms.py
```diff
@@ -160,103 +523,173 @@ def findatoms(observation):
...
     return atoms
 
+# --------------
+
 
-def group_proc(group):
-    '''Process observations within groups'''
+def group_proc(group, sel_obs_class=['SCIENCE', 'PROGCAL', 'PARTNERCAL', 'ACQ', 'ACQCAL', 'DAYCAL'],
+               sel_obs_status=['PHASE_2', 'FOR_REVIEW', 'IN_REVIEW', 'FOR_ACTIVATION', 'ON_HOLD', 'READY',
+                               'ONGOING', 'OBSERVED', 'INACTIVE'], fid=sys.stdout, wb=None, verbose=False):
```
sraaphorst, line 575: `ObservationClass` and `ObservationStatus` are defined in the mini-model, so I suspect we will want to move this to use those instead of `str` in the final version of this.

odb_extractor_atoms.py
```diff
@@ -160,103 +523,173 @@ def findatoms(observation):
...
+            print('', file=fid)
 
     return
 
 
-def prog_proc(program):
-    '''Process top-level of program'''
+def prog_proc(program, sel_obs_class=['SCIENCE', 'PROGCAL', 'PARTNERCAL', 'ACQ', 'ACQCAL', 'DAYCAL'], \
+              sel_obs_status=['PHASE_2', 'FOR_REVIEW', 'IN_REVIEW', 'FOR_ACTIVATION', 'ON_HOLD', 'READY', \
+                              'ONGOING', 'OBSERVED', 'INACTIVE'], fid=sys.stdout, xls=None, verbose=False):
```
sraaphorst, line 643: Same as above.

odb_extractor_atoms.py
```diff
@@ -288,7 +721,7 @@ def printseq(sequence, comment='', csv=False, path=''):
         if 'GMOS' in inst:
             coadds = '1'
```
sraaphorst, line 722: This is more a question of personal curiosity: I often see coadds represented as a string. Is there a reason for this? Aren't they always an integer?

odb_extractor_atoms.py
```diff
@@ -348,7 +781,7 @@ def seqxlsx(sequence, comment='', path=''):
...
             # convert wavelength to microns
-    #             wavelength = '{:5.3f}'.format(float(wavelength) / 1000.)
+#             wavelength = '{:5.3f}'.format(float(wavelength) / 1000.)
         else:
             coadds = step['observe:coadds']
         data.append(int(coadds))
```
sraaphorst, line 787: Never mind: I got my answer here. 👍 
I think it makes more sense to just work with `int` here:
```
coadds = 1 if 'GMOS' in inst else int(step['observe:coadds'])
```
(or keep it as the structured `if` `else` if we are going to need the wavelength conversion).

odb_extractor_atoms.py
```diff
@@ -392,7 +825,7 @@ def readseq(file, path):
     # Read and parse csv file: first line is a comment, second has column headings
     nline = 0
     for line in f:
```
sraaphorst, line 827: This should probably be changed to:
```
for nline, line in enumerate(f):
```
and then we can eliminate the initialization and management of `nline`.

odb_extractor_atoms.py
```diff
@@ -445,340 +878,157 @@ def xlsxseq(file, path):
...
+                    if verbose:
+                        print('QA states: ', qastates)
+                        print('Classes: ', classes)
+                        print('Guiding: ', guiding)
+                    atoms[-1]['guide_state'] = any(guiding)
+                    atoms[-1]['wavelength'] = sequence['wavelength'][-1]
+                    atoms[-1]['required_resources']['inst'] = sequence['inst'][-1]
+                    atoms[-1]['required_resources']['filter'] = sequence['filter'][-1]
+                    atoms[-1]['required_resources']['disperser'] = sequence['disperser'][-1]
+                    atoms[-1]['required_resources']['fpu'] = sequence['fpu'][-1]
```
sraaphorst, line 955: This pattern seems to be coming up repeatedly, which suggests to me that we need to factor it out.

odb_extractor_atoms.py
```diff
@@ -445,340 +878,157 @@ def xlsxseq(file, path):
...
+                    # Print basic atom info
+                    print(" \t exec_time: {:7.2f}, prog_time: {:7.2f}, part_time: {:7.2f}, guide_state: {}".
+                          format(atoms[-1]['exec_time'], atoms[-1]['prog_time'], atoms[-1]['part_time'],
+                                 atoms[-1]['guide_state']))
+                natom += 1
+                atoms.append({'id': natom, 'exec_time': 0.0, 'prog_time': 0.0, 'part_time': 0.0,
+                              'class': 'NONE', 'observed': False, 'qa_state': 'NONE', 'guide_state': False,
+                              'wavelength': 0.0,
+                              'required_resources': {'inst': 'NONE', 'filter': 'NONE', 'disperser': 'NONE',
+                                                     'fpu': 'NONE'}})
```
sraaphorst, line 966: (Same, include with above comment.)

sraaphorst:
Sorry for the bombardment of comments. This all looks good and I feel like going through it, I have a much better understanding now (and every time I read your code, you know some Python library that I've never used, so it gives me the opportunity to learn new things).

Most of my concerns and suggestions are in the individual comments. To summarize the larger issues:

1. I think will want to move away from `str` and `dict` and move towards more structured model objects when we finalize this code.
2. There is some repetition here that should be refactored out, especially with the atom assignment and creation.
3. The processing of the sequence info with so many nested `if` statements makes me a bit uncomfortable. I wonder if there is a way we can divide up this functionality into smaller pieces. This is something we can think about later: it certainly isn't essential now.
4. Error handling / logging should be included in the final version.

sraaphorst:
Also, perhaps we can work with @stroncod to make some test cases for this? Given the complexity, that might be a useful exercise.

odb_extractor_atoms.py
stroncod: What exactly? There are different things in the lines the comment is marked: 
- q offsets 
- exposure times and coadds (This are not in the mini-model, we might want to ask Bryan if are needed after all or not)
- the mode thing I deleted for the refactoring in the OCSProvider
-  do_not_split flag is also pointless

odb_extractor_atoms.py
bryanmiller: We will need do_not_split in general. For now it is just used for GPI but we may find other cases where it will be used.

sraaphorst: I'll mark this as resolved, then, with that understanding.

odb_extractor_atoms.py
sraaphorst: As Bryan pointed out, this logic will grow and we will need `do_not_split` for other things. I still recommend we move it to where it is first assigned if possible instead of at the beginning of the function as it does not appear for a long time in and that is slightly confusing, but this can evolve as we modify it for the ODB program provider, so I am marking this resolved.

PR #64, opened 2022-02-08
SCHED-93: Atom creation update

stroncod:
Update to our OCS Provider from the changes at the script #62

### DIFF
--- a/api/ocs/__init__.py
+++ b/api/ocs/__init__.py
@@ -58,6 +58,7 @@ class OcsProgramProvider(ProgramProvider):
 
     GPI_FILTER_WAVELENGTHS = {'Y': 1.05, 'J': 1.25, 'H': 1.65, 'K1': 2.05, 'K2': 2.25}
     NIFS_FILTER_WAVELENGTHS = {'ZJ': 1.05, 'JH': 1.25, 'HK': 2.20}
+    OBSERVE_TYPES = ['FLAT', 'ARC', 'DARK', 'BIAS'] 
 
     # We contain private classes with static members for the keys in the associative
     # arrays in order to have this information defined at the top-level once.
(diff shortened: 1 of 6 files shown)


sraaphorst (commented):
I think this is good but I need a bit more time to look it over.

api/ocs/__init__.py
```diff
@@ -372,78 +478,248 @@ def parse_nonsidereal_target(data: dict) -> NonsiderealTarget:
...
         """
-        n_steps = len(sequence)
-        n_abba = 0
+
+        def guide_state(step: dict) -> bool:
+            return any('guideWith' in key and guide == 'guide' for key, guide in step.items())
+        
+        def select_qastate(states: List[QAState]) -> QAState:
+            # Precedence order for observation classes.
+            qastate_order = [QAState.NONE, QAState.UNDEFINED, QAState.FAIL, QAState.USABLE, QAState.PASS]
```
sraaphorst, line 616: We don't need this (and furthermore, again, this list is being created every time the function is called): the `QAState` `enum` is actually an `IntEnum` that has this exact ordering on it, so we can just compare directly.

api/ocs/__init__.py
```diff
@@ -372,78 +478,248 @@ def parse_nonsidereal_target(data: dict) -> NonsiderealTarget:
...
+            return next((s for s in qastate_order if s in states), QAState.NONE)
```
sraaphorst, line 617: I think we just need:
```
return min(states, default=QAState.None)
```

api/ocs/__init__.py
```diff
@@ -372,78 +478,248 @@ def parse_nonsidereal_target(data: dict) -> NonsiderealTarget:
...
 
-            # Patterns:
-            # AB
-            # ABBA
-            if q is not None and n_steps >= 4 and n_steps - atom_id > 3 and n_abba == 0:
-                if (q == float(sequence[atom_id + 3][offset_q]) and
-                        q != float(sequence[atom_id + 1][offset_q]) and
-                        float(sequence[atom_id + 1][offset_q]) == float(sequence[atom_id + 2][offset_q])):
-                    n_abba = 3
+            if step[OcsProgramProvider._AtomKeys.OBSERVE_TYPE].upper() not in ['FLAT', 'ARC', 'DARK', 'BIAS']:
```
sraaphorst, line 667: We have this list of:
```
['FLAT' , 'ARC', 'DARK', 'BIAS']
```
in multiple places. We should move it out of the methods and into the class as a class variable.

api/ocs/__init__.py
```diff
@@ -492,7 +768,12 @@ def parse_observation(data: dict, num: int) -> Observation:
         qa_states = [QAState[log_entry[OcsProgramProvider._ObsKeys.QASTATE].upper()] for log_entry in
                      data[OcsProgramProvider._ObsKeys.LOG]]
 
+        
         atoms = OcsProgramProvider.parse_atoms(data[OcsProgramProvider._ObsKeys.SEQUENCE], qa_states)
+        #print(f'OBSERVATION_BASIC-{num}')
+        #print(atoms)
+        #input()
```
sraaphorst, line 775: Delete these comments. 😄

sraaphorst (changes requested):
A few small changes but otherwise, let's get this merged.

PR #65, opened 2022-02-14
SCHED-94: Readdition of InstConfig.

sraaphorst:
Readding the `InstConfig` class. We need to decide if we are storing all `Atom` resources in a single set or partitioning into an `InstConfig` class as demonstrated here.

NOTE: This PR is not YET ready for merging. If we go with this approach, we will have to add methods to some mini-model classes. This is for DISCUSSION purposes only currently.

### DIFF
--- a/common/minimodel/__init__.py
+++ b/common/minimodel/__init__.py
@@ -662,6 +662,7 @@ class Observation:
     def wavelengths(self) -> Set[float]:
         """
         The set of wavelengths included in the sequence.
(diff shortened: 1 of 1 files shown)


sraaphorst:
We have decided not to go with this approach. We are going to go with the original approach of simply using a `Set[Resource]` at the `Atom` level.

sraaphorst:
The wavelength fix has been migrated to:
https://github.com/gemini-hlsw/Scheduler/pull/70

PR #68, opened 2022-02-15
SCHED-98: Fixing test cases and bug in mini-model and API.

sraaphorst:
While running the test cases, which were suddenly failing, I found a bug in the way that observations were collected in the mini-model, and in how wavelengths were handled in general.


sraaphorst:
Force merging this as it is essential for more debugging.

PR #71, opened 2022-02-15
Sched 88: Clean up 

stroncod:


api/ocs/test/test_ocs_api.py
```diff
@@ -11,8 +11,8 @@ def get_api_program() -> Program:
     """
     Load the GN-2022A-Q-999 program from the JSON file.
     """
-    path = os.path.join('..', 'data', 'GN-2022A-Q-999.json')
-
+    path = os.path.join('data', 'GN-2022A-Q-999.json')
+    print(os.getcwd())
```
sraaphorst, line 15: Probably want to get rid of this. 😄

api/ocs/test/test_ocs_api.py
```diff
@@ -498,7 +503,7 @@ def create_minimodel_program() -> Program:
             qa_state=QAState.NONE,
             guide_state=False,
             resources={gmosn},
-            wavelength={0.475}
+            wavelengths={0.475}
```
sraaphorst, line 506: I think we might have maybe tread on each others toes on these changes since we did similar things? I guess we'll find out during merge time!

old/selector/__init__.py
```diff
@@ -2,19 +2,19 @@
 
 import astropy.units.quantity
 
-from old_collector import Collector
-import old_collector.vskyutil as vs
-import old_collector.sb as sb
+from collector import Collector
+import collector.vskyutil as vs
+import collector.sb as sb
```
sraaphorst, line 7: Not sure if this will work? I mean, it doesn't matter if we can't run the old selector at this point. I'm just curious if this is going to try to run something in this collector or in the new one?

old/selector/__init__.py
```diff
@@ -2,19 +2,19 @@
...
 from common.structures.observation_status import ObservationStatus
 from common.structures.target import TargetTag
 from common.structures.too_type import ToOType
 from common.structures.obs_class import ObservationClass
-from old_selector.visibility import Visibility
-from old_selector.ranker import Ranker
-import old_selector.horizons as hz
+from selector.visibility import Visibility
+from selector.ranker import Ranker
+import selector.horizons as hz
```
sraaphorst, line 17: That probably answers my questions since these won't be in the old ones.

scripts/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+1. download_programs.py 
+
+Downloads program needed for the Scheduler to work. 
+DEFAULT_PROGRAMS variable specifies the list of programs the script is going to download
+Programs (the .json.gz file) needs to be on the `data/` for the Provider to work properly.
+
+2. odb.extractor_atoms.py
+
+Script that generates atom for OCS programs. This is Bryan experimental work and does not use current Minimodel structures.
```
sraaphorst, line 11: Great! I'd love so see us all add a bit more of this. (I know I'm terrible at it.)

sraaphorst (approved):
I think it you just get rid of the `print` statement in `test_ocs_api.py` file, this is what we need, so I'll approve it now but please get rid of that line. 😄

old/selector/__init__.py
stroncod: You are right. Probably following the namespace should be `old.collector`

PR #76, opened 2022-02-23
SCHED-97 (Part 2): Cleanup, efficiency, improvements, defaults

sraaphorst:
This PR continues the work of the pre-Collector and Collector by:
1. Improving the efficiency of visibility calculations by moving the `sun_alt_indices` to the pre-Collector.
2. Handling cases where an `Observation` has no `Constraints`.
3. Implementing the case where an `Observation`'s elevation constraint type is `ElevationConstraintType.NONE`.
4. Improves (and fixes) dexception handling and logging.
5. Fixes an issue where `Observation` timing window representation was calculated many times over.
6. Removes redundant comments.

There are remaining tasks for the Collector to be production-worthy, but these will be addressed in separate Jira tasks.

### DIFF
--- a/scripts/run_collector.py
+++ b/scripts/run_collector.py
@@ -1,23 +1,34 @@
 import os
 
 from api.ocs import read_ocs_zipfile, OcsProgramProvider
+from common.output import print_collector_info
 from collector import *
 
 if __name__ == '__main__':
-    logging.basicConfig(level=logging.INFO)
+    # Reduce logging to ERROR only to display the tqdm bars nicely.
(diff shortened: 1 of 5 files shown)


scripts/run_collector.py
```diff
@@ -1,14 +1,19 @@
 import os
 
 from api.ocs import read_ocs_zipfile, OcsProgramProvider
+from common.output import print_collector_info
```
stroncod, line 4: I assume a `print_selector_info` is going to be added or we can just put that inside the collector

stroncod (approved):

PR #84, opened 2022-03-25
Sched 114: Proper Motion for sidereal Targets

stroncod:
New approach based on Bryan's approach: 
https://gist.github.com/bryanmiller/09202047fba8697be9af43735fc4a87c

Test needs a review on precision to pass correctly

### DIFF
--- a/collector/test/test_proper_motion.py
+++ b/collector/test/test_proper_motion.py
@@ -23,5 +23,5 @@ def test_proper_motion(target, time):
     Test that the sun location is at the equator at J2000.
     """
     coord = Collector._calculate_proper_motion(target, time)
-    assert coord.ra.deg == 78.85740081
-    assert coord.dec.deg == 15.63008372
+    assert abs(coord.ra.deg - 78.85740081) < 1e-5
+    assert abs(coord.dec.deg - 15.63008372) < 1e-5
--- a/collector/test/test_proper_motion.py
+++ b/collector/test/test_proper_motion.py
@@ -1,0 +1,27 @@
+import pytest
+from astropy.time import Time
+from collector import Collector
+from common.minimodel import SiderealTarget
+
+
+@pytest.fixture
+def time():
+    return Time('2022-03-31 12:00:00', scale='utc')
+
+@pytest.fixture
+def target():
+    return SiderealTarget(None, None, None,
+                          ra=78.85743749999999,
+                          dec=15.630119444444444,
+                          pm_ra=-5.718,
+                          pm_dec=-5.781,
+                          epoch=2000.0)
+
+
+def test_proper_motion(target, time):
+    """
+    Test that the sun location is at the equator at J2000.
+    """
+    coord = Collector._calculate_proper_motion(target, time)
+    assert coord.ra.deg == 78.85740081
+    assert coord.dec.deg == 15.63008372
--- a/components/collector/__init__.py
+++ b/components/collector/__init__.py
@@ -243,7 +243,7 @@ class Collector(SchedulerComponent):
     # The number of milliarcsecs in a degree, for proper motion calculation.
     _MILLIARCSECS_PER_DEGREE: ClassVar[int] = 60*60*1000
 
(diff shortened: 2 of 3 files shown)


collector/__init__.py
```diff
@@ -244,7 +245,9 @@ class Collector(SchedulerComponent):
     _MAX_NIGHT_EVENT_TIME: ClassVar[Time] = Time('2100-01-01 00:00:00', format='iso', scale='utc')
 
     # The number of milliarcsecs in a degree, for proper motion calculation.
-    _MILLIARCSECS_PER_DEGREE: ClassVar[int] = 1296000000
+    _MILLIARCSECS_PER_DEGREE: ClassVar[int] = 60*60*1000
+
+    _EPOCH2TIME: ClassVar[Mapping[float, Time]] = {2000.0: Time('2000-01-01 00:00:00', format='iso', scale='utc'), }
```
sraaphorst, line 250: What are we going to do in the case of other epochs? There should be a place where, if we encounter another epoch value, we add it to this, no? I would think we could just add a line as per below.

sraaphorst: Incidentally, I'm not sure that this is right. If you look up J2000:

"January 1, 2000
The currently-used standard epoch "J2000" is defined by international agreement to be equivalent to: The Gregorian date January 1, 2000, at 12:00 TT (Terrestrial Time). The Julian date 2451545.0 TT (Terrestrial Time)."

collector/__init__.py
```diff
@@ -364,6 +367,21 @@ def _process_timing_windows(prog: Program, obs: Observation) -> List[Time]:
...
+    @staticmethod
+    def _calculate_proper_motion(target: Target, time: Time) -> SkyCoord:
+        """
+        Calculate the proper motion of a target.
+        """
+
+        pm_ra = target.pm_ra / Collector._MILLIARCSECS_PER_DEGREE
+        pm_dec = target.pm_dec / Collector._MILLIARCSECS_PER_DEGREE
+        
+        time_offsets = time - Collector._EPOCH2TIME[target.epoch]
```
sraaphorst, line 380: Instead of this, we should probably have something like:
```
epoch_time = Collector._EPOCH2TIME.getdefault(target.epoch, Time(target.epoch, format='jyear'))
time_offsets = time - epoch_time
```

This unfortunately requires us to be reliant on AstroPy, but I suspect we can make our own function using `datetime` to do the calculation. It looks like, from https://en.wikipedia.org/wiki/Epoch_(astronomy) that:
```
J = 2000 + (Julian date − 2451545.0) ÷ 365.25
```
where J is the epoch, so we can get the Julian date:
```
jd = (epoch - 2000)  * 365.25 + 2451545.0
```

There is some simple code to convert between Julian dates and `datetime` here, I think:
https://github.com/dannyzed/julian/blob/master/julian/julian.py

that would allow us to avoid AstroPy.

collector/__init__.py
```diff
@@ -364,6 +367,21 @@ def _process_timing_windows(prog: Program, obs: Observation) -> List[Time]:
...
+        new_ra = (target.ra + pm_ra * time_offsets.to(u.yr).value) * u.deg
+        new_dec = (target.dec + pm_dec * time_offsets.to(u.yr).value) * u.deg
+        return SkyCoord(new_ra, new_dec, frame='icrs', unit='deg')
```
sraaphorst, line 383: I'll make a separate Jira task to get `SkyCoord` out of here and convert the mini-model to use instead, in decimal degrees:
```
ra: npt.NDArray[float]
dec: npt.NDArray[float]
```

stroncod: The thing with getting rid of SkyCoord here is that the Collector uses `.separation` after returning the value. I think the angular distance method in horizons might be enough to replace but that is untested

sraaphorst: Ahhh... good call. I had forgotten about that. I see that you made a comment in SCHED-112 how we can get around this. Excellent. 👍

sraaphorst (approved):
👍 Looks great!

sraaphorst:
Looks like my merge did result in some conflicts. :-/ Sorry.

stroncod:
Push the fix that solves the merge conflicts

PR #94, opened 2022-05-05
SCHED-125: First structure for Optimizer

stroncod:
Playing around with a generic Optimizer structure so adding and testing new algorithms is easier in the (not so far) future.
PR is meant to be a discussion for now and not to be merged.


stroncod:
I think some metaclass programming might help us improve BaseOptimizer, right now is just a simple interface that forces some methods to be implemented but I think an 'schedule' common method between algorithms can be accomplish (as a driver to interact with the Plan)

common/minimodel/__init__.py
```diff
@@ -1002,3 +1002,18 @@ def total_used(self) -> timedelta:
...
+    def __init__(self, night_duration=10):
+        self._time_slots_left = night_duration 
+        self._groups = []
+        
+    def add_group(self, group):
+        self.groups.append(group)
+        self._time_slots_left-=1 # TODO: clearly a missrepresantion between time allocates and plan timeslots, a proper function is in place
+    
+    def is_full(self):
+        return self._time_slots_left == 0
```
sraaphorst, line 1019: I'm not quite sure how this is intended to work? I've been struggling to think of a name for the Scheduler output since the reinterpretation of nightly plans but plan seems as good a name as anything.

What I'm not clear about here is the significance of `night_duration` (number of time slots in a night?) and why adding a group reduces the number of time slots left by exactly one.

Where is the concept of multiple nights represented here?

components/optimizer/greedymax.py
```diff
@@ -0,0 +1,27 @@
...
+    def __init__(self, some_parameter=1):
+        # Parameters specifically for the GreedyMax optimizer should be go here
+        self.some_parameter = some_parameter
+        self._visits = []
+        self.selection = None
+    
+    def _run(self):
+        ...
+
+    def add(self, selection):
```
sraaphorst, line 18: What type will `selection` be? I'm curious how you're envisioning this.
I'm guessing we'll need groups and their scores as the selection information?

stroncod:
This need more cleanup (add typing and delete some comments) but it has SCHED-140 work that adds a Dummy optimizer that could replace GM for further work

common/minimodel/__init__.py
```diff
@@ -1008,3 +1013,52 @@ def total_used(self) -> timedelta:
...
     def observations(self) -> List[Observation]:
         return self.root_group.observations()
+
+
+class Plans:
+    """
+    A collection of Nightly Plan from a corresponding Site
+    """
+    def __init__(self, night_events):
+        self.durations = [len(night_events.times[idx]) for idx, _ in enumerate(night_events.time_grid)]
```
sraaphorst, line 1023: This seems like a really unnecessary way to do it?
We can just do:
```
self.durations = [len(times) for times in night_events.times]
```
or:
```
self.durations = list(map(len, night_events.times))
```

sraaphorst: I don't think there's a reason that we have to iterate over the `time_grid`?

common/minimodel/__init__.py
```diff
@@ -1008,3 +1013,52 @@ def total_used(self) -> timedelta:
...
+        self.start_times = [night_events.local_times[idx][0] for idx, _ in enumerate(night_events.time_grid)]
```
sraaphorst, line 1024:
```
self.start_times = [local_times[0] for local_times in night_events.local_times]
```

common/minimodel/__init__.py
```diff
@@ -1008,3 +1013,52 @@ def total_used(self) -> timedelta:
...
+        self.end_times = [night_events.local_times[idx][-1] for idx, _ in enumerate(night_events.time_grid)]
```
sraaphorst, line 1025: Same as above but `-1` instead of `0`.

common/minimodel/__init__.py
```diff
@@ -1008,3 +1013,52 @@ def total_used(self) -> timedelta:
...
+    def __iter__(self):
+        self.plans = []
+        return self
+
+    def __next__(self):
+        if len(self.plans) <= len(self.durations):
+            self.plans.append(Plan(next(iter(self.start_times)),
+                                   next(iter(self.end_times)),
+                                   slot_length=self.time_slot_length,
+                                   night_duration=next(iter(self.durations))))
```
sraaphorst, line 1037: I'm not clear why we are appending in `__next__`.
Why don't we set this up beforehand?
I don't think we should be mutating in an iterator method. It seems like a dangerous idea.

sraaphorst: If we iterate twice over, you'll be repeating work, by clearing out `plans` and then recreating it, no?

And if we have two iterators, it'll play havoc and the value will depend on who calls `next` first / how often.

common/minimodel/__init__.py
```diff
@@ -1008,3 +1013,52 @@ def total_used(self) -> timedelta:
...
+class Plan:
+    """
+    A 'plan' is a collection of nighly plans
+    """
+    def __init__(self, start, end, slot_length=1, night_duration=10):
+        self.start = start
+        self.end = end
+        self._time_slot_length = slot_length
+        self._time_slots_left = night_duration
+        self._visits = []
```
sraaphorst, line 1051: This could easily be made a `dataclass` to be much simpler.
Also, why is the time slot length private? Won't this be needed?

```
@dataclass
class Plan:
    start: datetime
    end: datetime
    _time_slot_length: timedelta = timedelta(minutes=1)
    _time_slots_left: int = 10  # I'm not sure why 10 is chosen here?

    def __post_init__(self):
        self._visits = []
```

common/minimodel/__init__.py
```diff
@@ -1008,3 +1013,52 @@ def total_used(self) -> timedelta:
 
     def observations(self) -> List[Observation]:
         return self.root_group.observations()
+
+
+class Plans:
+    """
+    A collection of Nightly Plan from a corresponding Site
+    """
+    def __init__(self, night_events):
```
sraaphorst, line 1022: I'm really pushing for us to include type hints to make this code more production-worthy so that warnings will be generated if we pass incorrect data types in, which is far too easy to do in Python.

```
def __init__(self, night_events: NightEvents):
```

common/minimodel/__init__.py
```diff
@@ -1008,3 +1013,52 @@ def total_used(self) -> timedelta:
...
+    A 'plan' is a collection of nighly plans
+    """
+    def __init__(self, start, end, slot_length=1, night_duration=10):
+        self.start = start
+        self.end = end
+        self._time_slot_length = slot_length
+        self._time_slots_left = night_duration
+        self._visits = []
+    
+    def _time2slots(self, time) -> int:
```
sraaphorst, line 1053: I'm guessing it's `time: timedelta` based on the `total_seconds` call below?

common/minimodel/__init__.py
```diff
@@ -1008,3 +1013,52 @@ def total_used(self) -> timedelta:
...
+        return ceil((time.total_seconds() / 60) / self._time_slot_length.value)
+
+    def add_group(self, group):
```
sraaphorst, line 1056:
```
def add_group(self, group: Group) -> NoReturn:
```

common/minimodel/__init__.py
```diff
@@ -1008,3 +1013,52 @@ def total_used(self) -> timedelta:
...
+        self._visits.append(group)
+        self._time_slots_left -= self._time2slots(group[0].total_used())
+    
+    def is_full(self):
+        return self._time_slots_left <= 0
```
sraaphorst, line 1061: If `self._time_slots_left  < 0`, wouldn't this be a problem?

components/optimizer/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+from common.minimodel import Plans
+
+
+class Optimizer:
+    """
+    Entrypoint to interact with an BaseOptimizer object.
+    All algorithms need to follow the same structure to create a Plan
+    """
+
+    def __init__(self, selection, algorithm=GreedyMax(some_parameter=1)):
```
sraaphorst, line 11: We should talk about what the `selection` is.
I think I see below that it is a:
```
Dict[GroupID, Tuple[Group, GroupInfo]]
```
?

That seems good to me, but I am assuming that you will also need more information, such as `TargetInfo`, no?

components/optimizer/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+        self.algorithm = algorithm.add(selection)
+    
+    def schedule(self, night_events) -> Plans:
+        # TODO: This forces to make Plans for each site SEPARATELY.
+        # This would create and issue with OR groups as observations can be schedule 
+        # on different sites but not twice. Old GM had some check-ins to handle this 
+        # cases but right now we don't have OR groups implemented.
```
sraaphorst, line 18: We should talk about this and see what we need to do. I had the impression that a NightlyPlan would be site-specific, but a collection of plans would include nightly plans for all sites. If this is going to be a design problem, though, we will have to do something else.

components/optimizer/base.py
```diff
@@ -0,0 +1,40 @@
+import abc
+from common.minimodel import Group, Plan
+from components.selector import GroupInfo
+from typing import Dict
+
+Selection = Dict[Group, GroupInfo]
```
sraaphorst, line 6: LOL just saw this here, which contradicts my comment above, but what about L169 in the `run_scheduler.py` script below? I think maybe we should be looking up by `GroupID` and getting `Tuple[Group, GroupInfo]`? We may not be able to guarantee that a `Group` can be hashed.

components/optimizer/base.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class BaseOptimizer(metaclass=abc.ABCMeta):
```
sraaphorst, line 9: I'm not sure why you're using a `metaclass` here instead of inherting from `ABC`:

https://stackoverflow.com/questions/33335005/is-there-any-difference-between-using-abc-vs-abcmeta

components/optimizer/base.py
```diff
@@ -0,0 +1,40 @@
...
+    @classmethod
+    def __subclasshook__(cls, subclass):
+        return (hasattr(subclass, 'schedule') and
+                callable(subclass.schedule) and
+                hasattr(subclass, 'add') and
+                callable(subclass.add) and
+                hasattr(subclass, 'get_visits') and
+                callable(subclass.get_visits) and
+                hasattr(subclass, '_run') and
+                callable(subclass, '_run'))
```
sraaphorst, line 28: And maybe this is something that you know that I don't, because I haven't worked with `__subclasshook__` before and had to look it up, but why don't we just use `@abstractmethod`s for these? Wouldn't we want them all to have the same type, and have the type specified?

components/optimizer/base.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @property
+    def get_visits(self):
+        return self._visits
```
sraaphorst, line 32: I'm not entirely sure that this is the correct use of `@property`.

https://www.programiz.com/python-programming/property

This should probably just be called:
```
@property
def visits(self):
    return self._visits
```
but in this case, it really doesn't serve a purpose, because you're not doing any checking on `self._visits` and you're returning a reference to a list, so if anyone accesses it, they can change it and the changes will be propagated back to the `BaseOptimizer`.

At this point, you may as well just have `self.visits` and not make it "Python private" and allow users to access it directly.

If the intention is to prevent modification, you should have something like:

```
from copy import copy, deepcopy  # Not sure which one you want
...
@property
def visits(self):
    # One of these two:
    return copy(self._visits)
    return deepcopy(self._visits)
```

components/optimizer/dummy.py
```diff
@@ -0,0 +1,31 @@
...
+        random.seed(seed)
+
+    def _run(self, plan: Plan):
+        """
+        Gives a random group/observation to add to plan
+        """
+        ran_group = random.choice(list(self.selection.values()))
+        # TODO: A useful thing might be transforming the total times to time slots
+        # to reduce runtime calculations.
+        group_length = plan._time2slots(ran_group[0].total_used())
```
sraaphorst, line 19: Again, not clear why you're accessing `ran_group[0]`?

components/optimizer/dummy.py
```diff
@@ -0,0 +1,31 @@
...
+        else:
+            # TODO: DO a partial insert
+            # Spliting groups is not yet implemented
+            plan.add_group(ran_group)
+        print(f'slots left in the plan: {plan.time_slots_left()}')
+        
+    def add(self, selection: Selection):
+        # Preparation for the optimizer i.e create chromosomes, etc.
+        self.selection = selection
+        return self
```
sraaphorst, line 31: I'm not opposed to this, but it is a bit strange... I'm not sure why you'd return `self` here instead of just passing `selection` to the `schedule` method? It doesn't seem like you need this, and if you want to run again using the same Optimizer but the data has changed, it seems like it should be passed into `optimizer.schedule`.

scripts/run_scheduler.py
```diff
@@ -4,7 +4,9 @@
 from api.programprovider.ocs import read_ocs_zipfile, OcsProgramProvider
 from common.output import print_collector_info
 from components.collector import *
+from components.optimizer.dummy import Dummy
```
sraaphorst, line 7: LOL maybe we should call this `DummyOptimizer`? 😅

scripts/run_scheduler.py
```diff
@@ -158,9 +160,24 @@
...
     #    if program.id == 'GS-2018B-Q-101':
     #        atoms_to_sheet(program)
+
+    # "Selection" format for optimzer
+    # Dict[GroupID, Tuple[Group, GroupInfo]
+    selection = {}
+    for program_id, group_info_map in results.items():
+        for group_id, group_info in group_info_map.items():
+            group = selector.get_group(group_id)
+            selection[group_id] = (group, group_info)  # TODO: We might consider just add group info to group?
```
sraaphorst, line 175: This should probably be moved directly to the Selector. I can do that, but leave it here now so you can keep working on this until I've done that.

sraaphorst (commented):
Some things for discussion. We should talk about this more tomorrow (Wednesday).

common/minimodel/__init__.py
sraaphorst: Oh, one more concern: if a `Plan` is a collection of nightly plans, the number of time slots for each night will be different. So how will `_time_slots_left` work?

common/minimodel/__init__.py
```diff
@@ -1008,3 +1013,52 @@ def total_used(self) -> timedelta:
...
+        self._time_slot_length = slot_length
+        self._time_slots_left = night_duration
+        self._visits = []
+    
+    def _time2slots(self, time) -> int:
+        return ceil((time.total_seconds() / 60) / self._time_slot_length.value)
+
+    def add_group(self, group):
+        self._visits.append(group)
+        self._time_slots_left -= self._time2slots(group[0].total_used())
```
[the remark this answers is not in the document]
stroncod: `group` might be a bit confusing  name convention as the parameter is not a Group object but a pair `[Group, GroupInfo]`. So that access the Group object in fact

components/optimizer/base.py
stroncod: Fixed

stroncod: Fixed

components/optimizer/base.py
stroncod: So what I wanted to force was the subclasses to implement a self._visits member. This was removed because visits are now saved in the plan. So no need for this bad code

common/minimodel/__init__.py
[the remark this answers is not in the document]
sraaphorst: Got it. I see it now with the type annotation and it makes sense to me now.
Let's talk about the concept of nightly plans when we meet.

common/minimodel/__init__.py
```diff
@@ -1008,3 +1014,56 @@ def total_used(self) -> timedelta:
 
     def observations(self) -> List[Observation]:
         return self.root_group.observations()
+
+
+class Plan:
```
sraaphorst, line 1019: Just for discussion:

My thought was that we would have simple types for the plan and the optimizer would handle the timing and assembly of the data. Something like this:

```
@dataclass
class PlanEntry:
    start_time: datetime  # Unsure if this or something else
    obs_id: ObservationID
    atom_start_idx: int
    atom_end_idx: int

NightPlan = List[PlanEntry]
SitePlans = Mapping[NightIndex, NightPlan]
Plans = Mapping[Site, SitePlans]
```

Something like that.

sraaphorst: (That was just my thinking... not saying we have to go in that direction.)

scripts/run_scheduler.py
sraaphorst, line 175: I did think about adding `GroupInfo` to `Group` (same with `TargetInfo` to `Target`), but I would like to have the mini-model treated as immutably as possible for multiple reasons: it might need some restructuring, but we want to be able to have various validations / simulations - possibly with different time slot lengths or dates or what not - using the same mini-model data. If we include the calculated info into that, since we want to share that mini-model data, we could run into one thread overwriting another's info.

common/minimodel/__init__.py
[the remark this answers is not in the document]
sraaphorst: This also limits us in that we will be adding visits sequentially to the plan, which may cause problems. I don't think that is the behavior we want, since GreedyMax is not always going to schedule from start to end, and any other optimizer (if we end up writing another non-dummy one) will likely not schedule from start to end either.

This is why I was thinking of the simpler implementation of the concept of plans as per above.

components/optimizer/dummy.py
```diff
@@ -0,0 +1,75 @@
...
+    
+    def add(self, group: Group, plans: Plans) -> bool:
+        """
+        Add a group to a Plan
+        This is called when a new group is added to the program
+        """                        
+        # TODO: Missing different logic for different AND/OR GROUPS
+        # Add method should handle those
+        for observation in group.group.observations():
```
sraaphorst, line 61: This has me confused... If `group` is a `Group` as per the method type hints, it shouldn't have a member named `group`?

components/optimizer/greedymax.py
```diff
@@ -0,0 +1,59 @@
+from .base import BaseOptimizer
+from typing import Dict
+from common.minimodel.group import Group
+from common.minimodel.plan import Plan
+from components.selector import GroupInfo
+
+Selection = Dict[Group, GroupInfo]
```
sraaphorst, line 7: I think we should get rid of this since now we have an actual defined `Selection` in `common.calculations`.

components/optimizer/greedymax.py
```diff
@@ -0,0 +1,59 @@
...
+        return max(i.scores[0][0] for i in info)  # This just plain wrong and for demo purposes
+ 
+    def add(self, max_group: Group, plan: Plan):
+        """
+        Insert the group into the plan.
+        This might be general for all optimizers but for GreedyMax we need to
+        move things around as we insert (also splitting). It could be force to implement by
+        the abstract class
+        """
+        plan.add_group(max_group)
```
sraaphorst, line 36: Maybe update this since you changed the method signature in `Plan`?

components/optimizer/greedymax.py
```diff
@@ -0,0 +1,59 @@
...
+        scheduled = False
+
+        while not scheduled:
+            iter += 1
+
+            max_group = self._find_max_group()
+            if self._insert(max_group, plan):
+                scheduled = True
+            else:
+                pass
```
sraaphorst, line 54: This loop looks suspicious... I know this is kind of guidance code, but still.. let's try not to make Bryan confused by the new structure.

components/optimizer/greedymax.py
stroncod: That greedymax.py file is not being used and probably be replaced with Bryan's work

components/optimizer/base.py
```diff
@@ -0,0 +1,32 @@
...
+from typing import Mapping, Tuple, List
+
+
+class BaseOptimizer(ABC):
+    """
+    Base class for all Optimizer components.
+    Each optimizing algorithm needs to implement the following methods:
+
+    schedule: methods that triggers the formation of the plan
+    get_visits: methods that updates the plan
```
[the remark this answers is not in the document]
stroncod: It is

components/optimizer/dummy.py
```diff
@@ -0,0 +1,75 @@
...
+                print('group not added')
+        
+    def setup(self, programInfo: Mapping[ProgramID, ProgramInfo]):
+        """
+        Preparation for the optimizer i.e create chromosomes, etc.
+        """
+        self.groups = []
+        for p in programInfo.values():
+            self.groups.extend([g for g in p.group_data.values() if g.group.is_observation_group()])
+        return self
```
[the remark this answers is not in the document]
stroncod: Ok so it is being utilize in this line: 
`self.algorithm = algorithm.setup(selection.program_info)` 
If `return self` is removed then algorithm is set to None.

sraaphorst (approved):
Can you confirm the changes were made before merge?

components/optimizer/greedymax.py
sraaphorst: Ah... okay. I mistakenly thought it was a blueprint to guide him. Let's make sure he knows that this file is outdated.

PR #95, opened 2022-05-13
SCHED-126: Interface simplification.

sraaphorst:
Make everything available through the `Selector` so that we don't have to get certain things from the `Collector` and the `Selector`. The previous `Collector` methods are now accessed through the `Selector` and the base target information retrieval has been moved to the mini-model for simplicity so you can get an `Observation`'s base `Target` directly from the `Observation` instead of going through the `Collector`.


sraaphorst:
Merging for Bryan's use.

PR #96, opened 2022-05-18
Sched 95: Special Handling Cases

stroncod:
An implementation on how we can handling special cases to structures on the minimodel without modifying the base structures. This could be extended to handle more telescopes.

### DIFF
--- a/common/decorators/__init__.py
+++ b/common/decorators/__init__.py
@@ -1,6 +1,5 @@
 
 import numpy as np
-from datetime import timedelta
 
 def scalar_input(func):
     """
@@ -17,15 +16,3 @@ def scalar_input(func):
            
         return np.squeeze(func(*args, **kwargs))
     return wrapper
-
-
-def with_igrins_cal(func):
-    def add_calibration(self):
-        if ('IGRINS' in [res.id for res in self.required_resources()] and self.partner_used() > 0):
-            return func(self) + timedelta(seconds=(1 / 6))
-        return func(self)
-    return add_calibration
-
-"""
-def gemini(func)
-"""
--- a/common/decorators/__init__.py
+++ b/common/decorators/__init__.py
@@ -1,6 +1,6 @@
 
 import numpy as np
-from datetime import timedelta 
+from datetime import timedelta
 
 def scalar_input(func):
     """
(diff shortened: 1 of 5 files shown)


common/decorators/__init__.py
```diff
@@ -16,3 +17,15 @@ def wrapper(*args, **kwargs):
            
         return np.squeeze(func(*args, **kwargs))
     return wrapper
+
+
+def with_igrins_cal(func):
+    def add_calibration(self):
+        if ('IGRINS' in [res.id for res in self.required_resources()] and self.partner_used() > 0):
```
sraaphorst, line 24: I'd prefer us to not pull out resource IDs directly and compare strings. We should avoid comparing strings like this whenever possible.

In `GeminiProperties`, we have an `Instruments` `Enum` that contains an entry for `IGRINS`. Instead, this should be:

```
if GeminiProperties.Instruments.IGRINS in self.required_resources() and self.partner_used() > 0:
```

Also, no parentheses needed.

sraaphorst, line 24: I'd prefer us to not pull out resource IDs directly and compare strings. We should avoid comparing strings like this whenever possible.

In `GeminiProperties`, we have an `Instruments` `Enum` that contains an entry for `IGRINS`. Instead, this should be:

```
if GeminiProperties.Instruments.IGRINS in self.required_resources() and self.partner_used() > 0:
```

Also, no parentheses needed.

common/minimodel/__init__.py
```diff
@@ -790,6 +791,15 @@ def __eq__(self, other: 'Observation') -> bool:
                 dict((k, v) for k, v in other.__dict__.items() if k != 'sequence'))
 
 
+class GeminiObservation(Observation):
+    """
+    A Gemini-specific extension of the Observation class.
+    """
+    @with_igrins_cal
+    def total_used(self) -> timedelta:
+        return super().total_used()
```
sraaphorst, line 800: This looks good to me.

sraaphorst, line 800: This looks good to me.

common/decorators/__init__.py
```diff
@@ -16,3 +17,15 @@ def wrapper(*args, **kwargs):
...
+            return func(self) + timedelta(seconds=(1 / 6))
+        return func(self)
+    return add_calibration
+
+"""
+def gemini(func)
+"""
```
sraaphorst, line 31: Just curious what this is about?

sraaphorst (changes requested):
Let's change the Resource thing as I don't think unless absolutely necessary (and hopefully it won't be), we shouldn't be working directly with Resource IDs or making str comparisons, and then we can merge this! The decorator idea works for me.

sraaphorst (approved):
👍

PR #97, opened 2022-05-31
SCHED-141: Filter out unschedulable groups

sraaphorst:
Now info for groups that are unschedulable during the time period under consideration are filtered out of the `Selector.select` output.


stroncod (approved):
I'll go and merge this to finish #94

PR #100, opened 2022-06-14
Sched 137: Hookup of process management with Scheduler code

stroncod:
Not to merge! (yet)


common/minimodel/site.py
```diff
@@ -35,5 +35,7 @@ def __init__(self, site_name: str, coordinate_center: str, astropy_lookup: Optio
             msg = f'Unknown time zone lookup: {timezone_info}.'
             raise ValueError(e, msg)
 
+    def __repr__(self) -> str:
+        return f"SiteInformation('{self.name}', '{self.coordinate_center}')"
```
sraaphorst, line 39: We should get rid of this now since `SiteInformation` is gone and it doesn't make sense any more.

process_manager/manager.py
```diff
@@ -23,6 +31,30 @@ def schedule_with_runner(self, task: SchedulerTask, mode: TaskType):
...
+        
+        asyncio.get_event_loop().add_signal_handler(signal.SIGINT, shutdown)
+
+        while not done.is_set():
+            self.add_task(datetime.now(), scheduler, mode)
+            if period == 0:
+                # random case #
+                await asyncio.sleep(randint(1, 10))
+            else:
+                await asyncio.sleep(period)
```
sraaphorst, line 57: Do we still need this random sleeping? I thought that was only for demo purposes?

process_manager/task.py
```diff
@@ -18,14 +18,12 @@ class SchedulerTask:
...
                  target: callable,
-                 timeout: int = DEFAULT_TIMEOUT) -> None:
+                 timeout: int) -> None:
 
         self.start_time = start_time
-        self.end_time = end_time
-        self.job_id = hash((self.start_time, self.end_time))
-        self.timeout = timeout
+        self.job_id = hash((self.start_time, timeout))
+        self.timeout = timeout if timeout is not None else DEFAULT_TIMEOUT
```
sraaphorst, line 26: Make `timeout` have type `Optional[int]`, or why not just have:
```
timeout: int = DEFAULT_TIMEOUT
```
like you did before?

scripts/run_with_manager.py
```diff
@@ -0,0 +1,41 @@
...
+import asyncio
+import logging
+from astropy.time import Time, TimeDelta
+import astropy.units as u
+from process_manager import TaskType, ProcessManager, Scheduler
+from process_manager import SchedulerConfig, CollectorConfig, SelectorConfig
+from api.observatory.gemini import GeminiProperties
+from common.minimodel import ALL_SITES, Semester, SemesterHalf, ProgramTypes, ObservationClass
+
+logging.basicConfig(level=logging.INFO)
```
sraaphorst, line 11: We should probably still have this in:
```
if __name__ == '__main__':
```
since it's Python convention to make all executables check for that before running.

sraaphorst (changes requested):
Just a question and a couple minor tweaks.
Looks really good! 👍

common/minimodel/site.py
stroncod: Got mixed up in the rebase

process_manager/manager.py
stroncod: Right now is looping from task to task.  It should be deleted in the future work

sraaphorst (approved):
Good stuff. 👍

PR #103, opened 2022-06-16
SCHED-147: Fixed issue where fpu had a value of None due to missing key in sequence.

sraaphorst:
This might be an error that we should check with in science.
In this case, the observation is GN-2018B-Q-103-19 and the failure is in `OcsProgramProvider._parse_instrument_configuration` here (L506):
```
            if instrument in OcsProgramProvider.FPU_FOR_INSTRUMENT:
                if OcsProgramProvider.FPU_FOR_INSTRUMENT[instrument] in data:
                    fpu = data[OcsProgramProvider.FPU_FOR_INSTRUMENT[instrument]]
                else:
                    # TODO: Might need to raise an exception here. Check code with science.
                    # TODO: An example of where this happens is for GN-2018B-Q-103-19.
                    fpu = None
```

It seems the key `instrument:fpu` is missing from the sequence data. I'm not sure if this is an error or not. Why would the key be missing instead of using `None` or `FPU_NONE`?

### DIFF
--- a/api/programprovider/ocs/__init__.py
+++ b/api/programprovider/ocs/__init__.py
@@ -501,8 +501,9 @@ class OcsProgramProvider(ProgramProvider):
                 if OcsProgramProvider.FPU_FOR_INSTRUMENT[instrument] in data:
                     fpu = data[OcsProgramProvider.FPU_FOR_INSTRUMENT[instrument]]
                 else:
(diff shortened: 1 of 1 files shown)


sraaphorst:
Possible issue / error reported to Slack #scheduler group and waiting for science to respond.

PR #104, opened 2022-06-16
HOTFIX: Fixing time information, type hints, calculations, and cleaning output

sraaphorst:
This fixes several things:

- Some of the typehints in the `Plans` code was incorrect.
- The `sky` library was not using `Moon.time_by_altitude` on moon set, so a triple was returned instead of the correct value.
- Times were not being wrapped correctly in `AstroPy``Time` and `TimeDelta` objects. This led them to have type `Quantity` instead of `Time` or `TimeDelta`, which was causing problems.
- Fix (possibly temporary) for FPUs that have a value of `None` during atom extraction.
- Remove tqdm, add Django (for now for GraphQL).
- Output fixes and commenting out uninformative logging info.

### DIFF
--- a/common/sky/events.py
+++ b/common/sky/events.py
@@ -7,19 +7,20 @@ from astropy.coordinates import Angle, EarthLocation
 from astropy.time import Time
 from pytz import timezone
 
+from common.types import TimeScalarOrNDArray
 from common.sky.constants import EQUAT_RAD
 from common.sky.moon import Moon
 from common.sky.sun import Sun
 from common.sky.utils import local_midnight_time
 
 
-def night_events(time: Time, location: EarthLocation, localtzone: timezone) -> \
-        Tuple[Time, Union[npt.NDArray[float], Time],
-              Union[npt.NDArray[float], Time],
-              Union[npt.NDArray[float], Time],
-              Union[npt.NDArray[float], Time],
-              Union[npt.NDArray[float], Time],
-              Union[npt.NDArray[float], Time]]:
+def night_events(time: Time, location: EarthLocation, localtzone: timezone) -> Tuple[Time,
+                                                                                     TimeScalarOrNDArray,
+                                                                                     TimeScalarOrNDArray,
+                                                                                     TimeScalarOrNDArray,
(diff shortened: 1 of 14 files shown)


common/sky/events.py
```diff
@@ -14,9 +14,12 @@
...
 def night_events(time: Time, location: EarthLocation, localtzone: timezone) -> \
-        Tuple[Time, Union[npt.NDArray[float], Time], Union[npt.NDArray[float], Time], Union[npt.NDArray[float], Time],
-              Union[npt.NDArray[float], Time], Union[npt.NDArray[float], Time],
-              Union[npt.NDArray[float], Tuple[Angle, Time, EarthLocation]]]:
+        Tuple[Time, Union[npt.NDArray[float], Time],
+              Union[npt.NDArray[float], Time],
+              Union[npt.NDArray[float], Time],
+              Union[npt.NDArray[float], Time],
+              Union[npt.NDArray[float], Time],
+              Union[npt.NDArray[float], Time]]:
```
stroncod, line 22: We might need an alias for this?

common/sky/test/test_events.py
```diff
@@ -19,5 +19,5 @@ def test_night_events(midnight, location):
     nptest.assert_almost_equal(even_12twi.jd, Time('2020-07-01 05:58:33.537').jd, decimal=2)
     nptest.assert_almost_equal(morn_12twi.jd, Time('2020-07-01 14:53:11.218').jd, decimal=2)
     nptest.assert_almost_equal(moonrise.jd, Time('2020-07-01 00:50:37.391').jd, decimal=2)
-    nptest.assert_almost_equal(moonset[1].jd, Time('2020-07-01 12:51:54.812').jd, decimal=2)
+    nptest.assert_almost_equal(moonset.jd, Time('2020-07-01 12:51:54.812').jd, decimal=2)
```
stroncod, line 22: Right, it makes way more sense now

stroncod (approved):

common/sky/events.py
sraaphorst: LOL probably a good idea. I'll add that.

PR #114, opened 2022-06-22
SCHED-160:Made the selector use Env for environmental data retrieval …

Davidwk123:
…instead of the Collector.

Also deleted older "get_actual_conditions_variant() method in the "/Scheduler/components/collector/__init__.py" path

### DIFF
--- a/components/collector/__init__.py
+++ b/components/collector/__init__.py
@@ -484,21 +484,3 @@ class Collector(SchedulerComponent):
                 Resource(id='H'),
                 Resource(id='J')
             }))
-
-    def get_actual_conditions_variant(self,
-                                      site: Site,
-                                      times: Time) -> Optional[Variant]:
-        """
-        Return the weather variant.
-        This should be site-based and time-based.
-        """
-        night_length = len(times)
-
-        return Variant(
-            iq=np.full(night_length, ImageQuality.IQ70),
-            cc=np.full(night_length, CloudCover.CC50),
-            wv=np.full(night_length, WaterVapor.WVANY),
-            wind_dir=Angle(np.full(night_length, 330.0), unit='deg'),
-            wind_sep=Angle(np.full(night_length, 40.0), unit='deg'),
-            wind_spd=Quantity(np.full(night_length, 5.0 * u.m / u.s))
-        )
--- a/components/selector/__init__.py
+++ b/components/selector/__init__.py
@@ -1,6 +1,6 @@
 import logging
 from dataclasses import dataclass
-from typing import Dict, FrozenSet, Set, Optional
+from typing import Dict, FrozenSet, Set, Optional, ClassVar
 
 import astropy.units as u
 import numpy as np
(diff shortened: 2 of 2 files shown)


components/collector/__init__.py
```diff
@@ -484,21 +484,3 @@ def available_resources(site: Site,
...
-        night_length = len(times)
-
-        return Variant(
-            iq=np.full(night_length, ImageQuality.IQ70),
-            cc=np.full(night_length, CloudCover.CC50),
-            wv=np.full(night_length, WaterVapor.WVANY),
-            wind_dir=Angle(np.full(night_length, 330.0), unit='deg'),
-            wind_sep=Angle(np.full(night_length, 40.0), unit='deg'),
-            wind_spd=Quantity(np.full(night_length, 5.0 * u.m / u.s))
-        )
```
sraaphorst, line 504: 😄 This was the next task I was going to give you. I don't know how VSC works, so I don't know if it tells you what imports you don't need any more, but there should be imports that are not being used any longer in this file. Could you remove them? (That was the second part of the next task.)

components/selector/__init__.py
```diff
@@ -27,7 +28,7 @@ class Selector(SchedulerComponent):
     information is statically determined.
     """
     collector: Collector
-
+    static_Env: ClassVar[Env] = Env()
```
sraaphorst, line 31: Also, not to be nitpicky, but Python has a very well-defined set of how naming conventions should work. (This is why I use PyCharm instead of MVC... it tells me all the warnings that violate Python standards and I don't think that MVC does.)

Just an FYI: variable names in Python (unless they're constants) are supposed to always be written in snake case, which is lowercase with underscores, so this should be called `static_env` instead of `static_Env`.

In fact, something I didn't mention, but we may as well go and make this variable "private" to indicate that it shouldn't be accessed outside of the class since we'll only need it in here and nowhere else. That means starting the name with an underscore.

So let's just declare it as:
```
_env: ClassVar[Env] = Env()
```
which is right.

components/selector/__init__.py
```diff
@@ -217,7 +218,7 @@ def _calculate_observation_group(self,
         night_events = self.collector.get_night_events(obs.site)
 
         for night_idx in ranker.night_indices:
-            actual_conditions = self.collector.get_actual_conditions_variant(obs.site, night_events.times[night_idx])
+            actual_conditions = Selector.static_Env.get_actual_conditions_variant(obs.site, night_events.times[night_idx])
```
sraaphorst, line 221: ...which would make this:
```
actual_conditions = Selector._env.get_actual_conditions_variant(obs.site, night_events.times[night_idx])
```

So, great job. You did it just right and more than I asked, so if you make these changes, we can merge.

sraaphorst (changes requested):
Good work. A few changes requested and then we can merge it and move on.

sraaphorst (approved):

PR #118, opened 2022-06-28
Sched 150: App structure

stroncod:


app/app.py
```diff
@@ -0,0 +1,39 @@
...
+import signal
+from datetime import datetime
+from .process_manager import ProcessManager, TaskType
+from .scheduler import Scheduler
+
+TASK_QUEUE = []
+
+
+class App:
+    def __init__(self, config):
```
sraaphorst, line 11: Can we say anything about the type of `config`?

app/config.py
```diff
@@ -0,0 +1,25 @@
+from dataclasses import dataclass
+from astropy.time import Time
+from astropy.units import Quantity
+
+
+@dataclass
+class CollectorConfig:
+    semesters: set
+    program_types: set
+    obs_classes: set
```
sraaphorst, line 10: We should make these `FrozenSet[Semester]`, etc.

app/config.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+@dataclass
+class SelectorConfig:
+    properties: type
```
sraaphorst, line 15: Can we say anything more than `type`?

app/config.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+@dataclass
+class SchedulerConfig:
+    start_time: Time
+    end_time: Time
+    time_slot_length: Quantity
+    sites: set
```
sraaphorst, line 23: `FrozenSet[Site]`

sraaphorst (commented):
This looks good to me, except for concerns about the type hints.
Did you leave them out for a reason?

sraaphorst (approved):

app/app.py
stroncod: This is address in #120

app/config.py
stroncod: This is removed in #120

app/config.py
stroncod: Same as above

PR #120, opened 2022-06-29
Sched 149: Add configuration manager to App

stroncod:


sraaphorst (approved):
I think this may answer my questions in the other PR if you're using `yaml` configs, but I'm not familiar with `OmegaConf` and will look it up tomorrow. I'll approve both the PRs, but if possible to add more type info (I'm not entirely sure), please do.

PR #123, opened 2022-07-04
HOTFIX: Fix environment.yml to work with pip dependencies

stroncod:
Las changes broke the env for the CI, this should fix it

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -14,4 +14,5 @@ dependencies:
   - coverage
   - requests
   - pandas
-  - omegaconf
+  - pip:
+    - omegaconf


environment.yml
```diff
@@ -14,4 +14,5 @@ dependencies:
   - coverage
   - requests
   - pandas
```
sraaphorst, line 16: The examples I've seen all have:
```
 - pip
 - pip:
    - omegaconf
    - etc
```
so maybe we should follow convention and put `pip` in there even though we can be pretty sure it will always be included.

sraaphorst (approved):

PR #125, opened 2022-07-06
SCHED-169: Made Env into a proper class

Davidwk123:
The commented code is my current work on the weather data dictionary

### DIFF
--- a/mock/environment/__init__.py
+++ b/mock/environment/__init__.py
@@ -1,5 +1,9 @@
 from typing import Optional
 
+import os
+import logging
+import bz2
+import pandas as pd
 import astropy.units as u
 import numpy as np
 from astropy.coordinates import Angle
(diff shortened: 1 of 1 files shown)


mock/environment/__init__.py
```diff
@@ -8,8 +12,36 @@
 
 from common.minimodel import Site, Variant, CloudCover, ImageQuality, WaterVapor
 
+logging.basicConfig(level=logging.INFO)
```
sraaphorst, line 15: This line should only be in the code where we have:
```
if __name__ == '__main__':
    # Code goes here
```
Don't worry about it now. If you set it up here, though, it will affect the whole Scheduler, which we don't want. We only want to set the logging level if the Env historical data is running separately.

mock/environment/__init__.py
```diff
@@ -8,8 +12,36 @@
...
+                with bz2.open(input_filename) as input_file:
+                        input_data = pd.read_pickle(input_file)
+                        self.site_data[site_lc] = input_data 
+                        #logging.info(self.site_data[site_lc]["Time_Stamp_UTC"].loc[150:200])
+                        logging.info(f'\t{len(self.site_data[site_lc].columns)} columns, {len(self.site_data[site_lc])} rows')
+                # self.site_data_by_night[site_lc] = self.night
+                # for night in self.site_data[site_lc]["Time_Stamp_UTC"]:
+                #     night_date = night.date()
+                #     #logging.info(night_date)
+                #     self.site_data_by_night[site_lc][night_date] = self.site_data[site_lc][0]
```
sraaphorst, line 39: If you get stuck here and need some help, let me know and I'll see if we can work through the process.

mock/environment/__init__.py
```diff
@@ -27,3 +59,9 @@ def get_actual_conditions_variant(self,
             wind_sep=Angle(np.full(night_length, 40.0), unit='deg'),
             wind_spd=Quantity(np.full(night_length, 5.0 * u.m / u.s))
         )
+    
+    def data_file_path(self, filename: str) -> str:
```
sraaphorst, line 63: For this, you can make is a static method since it doesn't depend on anything in the class.

You can mark it:
```
@staticmethod
def data_file_path(filename: str) -> str:
    # ...
```
(It lets you get rid of the `self` since you don't need anything from `self`.)

You can then change the call up there from the way you have it:
```
input_filename = self.data_file_path(f'{site_lc}_weather_data.pickle.bz2')
```

to:

```
input_filename = Env.data_file_path(f'{site_lc}_weather_data.pickle.bz2')
```

sraaphorst (changes requested):
Just a couple of changes, and then we can merge this.

mock/environment/__init__.py
sraaphorst: @stroncod `data_file_path` is a method that uses `os.path.join`. 😄

PR #126, opened 2022-07-06
SCHED-171: Added some methods to AndGroup and refactoring.

sraaphorst:
This adds:
* exec_time
* total_used
* instruments

to AndGroup. Also, relatively major refactoring on the `common.api.observatory` package and its functionality.

We now require an implementation of `common.api.observatory.ObservatoryProperties` to be plugged into the class so that the `ObservatoryProperties` can be used anywhere in the code base without a reference to them.


environment.yml
```diff
@@ -14,5 +14,6 @@ dependencies:
   - coverage
   - requests
   - pandas
+  - pip
```
stroncod, line 17: Thanks! I forgot to pushed that change in the last PR

scripts/run_scheduler.py
```diff
@@ -12,6 +13,7 @@
 
 if __name__ == '__main__':
     logging.basicConfig(level=logging.INFO)
+    ObservatoryProperties.set_properties(GeminiProperties)
```
stroncod, line 16: I like that, a decorator might be helpful too

stroncod (approved):
Just a couple of comments.

PR #128, opened 2022-07-08
Sched 170

Davidwk123:


mock/environment/__init__.py
```diff
@@ -16,34 +17,49 @@
 class Env:
    
     def __init__(self):
+        """
+        Stores weather data into a dictionary 
+        that can be indexed by site and date.
+        """
```
sraaphorst, line 23: Instead of reusing `'Time_Stamp_UTC'` over and over again, which makes it prone to typos, use a variable instead.

Same with :
```
day_difference = timedelta(hours=7)
```
so that we can change this easily.

mock/environment/__init__.py
```diff
@@ -16,34 +17,49 @@
...
         self.site_data = {}
```
sraaphorst, line 24: This confuses me, because we don't need the `site_data` stored in here.
We just need the `site_data_by_night`.

mock/environment/__init__.py
```diff
@@ -16,34 +17,49 @@
...
         self.site_data_by_night = {}
-        
+
         for site in Site:
             site_lc = site.name.lower()
```
sraaphorst, line 28: We only need `site_lc` once for the filename... the rest of the dictionary keys should just be `site`.

mock/environment/__init__.py
```diff
@@ -16,34 +17,49 @@
...
             input_filename = Env.data_file_path(f'{site_lc}_weather_data.pickle.bz2')
```
sraaphorst, line 29: Also, to make things simpler, I moved `data_file_path` into `__init__`. (In Python, you can have functions inside functions.) We'll never use it anywhere else, so we may as well put it here.

mock/environment/__init__.py
```diff
@@ -16,34 +17,49 @@
...
+
+                self.site_data_by_night[site_lc] = {}
+                local_site_data = self.site_data[site_lc].iterrows()
+                for _index, night in local_site_data:
+                    night_start_line = night
+                    night_date = night["Time_Stamp_UTC"].date()
+                    night_date = night_date.strftime('%Y-%m-%d') 
+                    night_list = [night_start_line]
+                    preivous_line = night_start_line
+                    _index2, current_line = next(local_site_data)
```
sraaphorst, line 48: Again, why _ in from of index2?

mock/environment/__init__.py
```diff
@@ -68,4 +84,5 @@ def data_file_path(filename: str) -> str:
 
 # if __name__ == '__main__':
 #     logging.basicConfig(level=logging.INFO)
-#     env = Env()
\ No newline at end of file
+#     env = Env()
+#     print(env.site_data_by_night['gs']["2014-01-02"])
```
sraaphorst, line 88: This is what I ultimately ended up with after cleaning things up a bit here. Your code did pretty much what we wanted it to do, but there was a little room for improvement. It might look a lot but it's actually quite minor. You were just doing a bit too much and a bit of unnecessary things in some places.

I suggest you take a quick look between our code and see how it differs because you were pretty much spot on except for some extra stuff  and some style issues:


```
import logging
import bz2
import os
from datetime import date, timedelta
from typing import Optional

import astropy.units as u
import numpy as np
import pandas as pd
from astropy.coordinates import Angle
from astropy.time import Time
from astropy.units import Quantity

from common.minimodel import Site, Variant, CloudCover, ImageQuality, WaterVapor


class Env:
    def __init__(self):
        """
        Stores weather data into a dictionary
        that can be indexed by site and date.
        """
        time_stamp = "Time_Stamp_UTC"
        day_difference = timedelta(hours=7)

        def data_file_path(filename: str) -> str:
            return os.path.join('..', '..', 'data', filename)

        # This is ultimately what we want.
        self.site_data_by_night = {}

        for site in Site:
            logging.info(f'Processing {site.name}')
            input_filename = data_file_path(f'{site.name.lower()}_weather_data.pickle.bz2')

            logging.info(f'\tReading {input_filename}')
            with bz2.open(input_filename) as input_file:
                input_data = pd.read_pickle(input_file)
                logging.info(f'\t\t{len(input_data.columns)} columns, {len(input_data)} rows')

            self.site_data_by_night[site] = {}
            local_site_data = input_data.iterrows()
            for index, night in local_site_data:
                night_start_line = night

                night_date = night[time_stamp].date()
                logging.info(f'\tProcessing UTC night of {night_date}')

                night_list = [night_start_line]
                preivous_line = night_start_line
                index2, current_line = next(local_site_data)

                while current_line[time_stamp] - preivous_line[time_stamp] < day_difference:
                    night_list.append(current_line)
                    preivous_line = current_line
                    try:
                        index3, current_line = next(local_site_data)
                    except StopIteration:
                        logging.info("End of data")
                        break

                self.site_data_by_night[site][night_date] = night_list

    def get_actual_conditions_variant(self,
                                      site: Site,
                                      times: Time) -> Optional[Variant]:
        """
        Return the weather variant.
        This should be site-based and time-based.
        """
        night_length = len(times)

        return Variant(
            iq=np.full(night_length, ImageQuality.IQ70),
            cc=np.full(night_length, CloudCover.CC50),
            wv=np.full(night_length, WaterVapor.WVANY),
            wind_dir=Angle(np.full(night_length, 330.0), unit='deg'),
            wind_sep=Angle(np.full(night_length, 40.0), unit='deg'),
            wind_spd=Quantity(np.full(night_length, 5.0 * u.m / u.s))
        )


if __name__ == '__main__':
    logging.basicConfig(level=logging.INFO)
    env = Env()

    # This logs way too much but is good for testing.
    # logging.info(env.site_data_by_night[Site.GS][date(2014, 1, 2)])
```

sraaphorst (changes requested):

mock/environment/__init__.py
```diff
@@ -16,34 +17,49 @@
...
-                #     night_date = night_date.strftime('%Y-%m-%d') 
-                #    # logging.info(night_date)
-                #     self.site_data_by_night[site_lc][night_date] = self.site_data[site_lc].iloc[count]
-                #     count += 1
-                #     print(self.site_data_by_night[site_lc][night_date])
-            else:
+
+                self.site_data_by_night[site_lc] = {}
+                local_site_data = self.site_data[site_lc].iterrows()
+                for _index, night in local_site_data:
```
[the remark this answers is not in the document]
Davidwk123: I thought the front index is used to show an unused variable

sraaphorst (approved):

mock/environment/__init__.py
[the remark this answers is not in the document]
sraaphorst: Oh, yeah, the pattern `_variable` is used to indicate that variables should be treated as private in a class, but at the same time, if you're not using a value, you should just call it `_` with nothing after it.

That's why most of the variables in classes are usually labeled:
```
self._variable
```
so that they're only accessible through methods or used in internal calculations.

PR #129, opened 2022-07-12
SCHED-171: Changed NaN/None/set values in Environment data.

Davidwk123:


sraaphorst (approved):
I think this looks great! Good work.

PR #134, opened 2022-07-18
SCHED-173:Reformatted Env class, deleted weather.py and fixed foward …

Davidwk123:
…filling data for each night


mock/environment/__init__.py
```diff
@@ -16,66 +16,120 @@
...
+                    while current_line[Env._time_stamp] - previous_line[Env._time_stamp] < Env._day_difference:
+                        night_list.append(current_line)
+                        previous_line = current_line
+                        try:
+                            index3, current_line = next(local_site_data)
+                            current_line["cc_band"] = Env._cc_band_to_float(current_line["cc_band"])
+                            current_line["iq_band"] /= 100
+                        except StopIteration:
+                            logging.info("End of data")
+                            break
```
sraaphorst, line 110: I have some concerns about this not working properly because you read in from the iterator three times and there's no guarantee that there will be three rows to read.

It could also be cleaned up since you do almost exactly the same thing three times (so it should be extracted to a function).

Let's worry about that later and make sure this works.

mock/environment/__init__.py
```diff
@@ -16,66 +16,120 @@
...
+                    cc_band = self.site_data_by_night[site][night].iloc[0]["cc_band"]
+                    starting_index = self.site_data_by_night[site][night].index[0]
+
+                    if pd.isna(iq_band) and pd.isna(cc_band):
+                        self.site_data_by_night[site][night].at[starting_index, "iq_band"] = 1.0
+                        self.site_data_by_night[site][night].at[starting_index, "cc_band"] = 1.0
+                    elif pd.isna(iq_band) and (~pd.isna(cc_band)):
+                        self.site_data_by_night[site][night].at[starting_index, "iq_band"] = 1.0
+                    elif (~pd.isna(iq_band)) and pd.isna(cc_band):
+                        self.site_data_by_night[site][night].at[starting_index, "cc_band"] = 1.0
```
sraaphorst, line 126: I'm not sure you understand what `~` does in Python. It means to invert all the bits and doesn't mean `not`. In Python, you just use the word `not` for not.

This could be simplified quite a bit:
```
if pd.isna(iq_band):
    self.site_data_by_night[site][night].at[starting_index, "iq_band"] = 1.0

if pd.isna(cc_band):
    self.site_data_by_night[site][night].at[starting_index, "cc_band"] = 1.0
```

You don't need four cases. You can just check `iq_band` and `cc_band` separately and handle them.

sraaphorst (changes requested):
If you could simplify the things I mentioned in my second comment, then we can check it in and make sure it's working. If everything is good, we can simplify the earlier code and make sure it doesn't cause any errors. 👍

sraaphorst (approved):

PR #138, opened 2022-07-20
SCHED-179: Bugfix in missing first row for each date and simplification.

sraaphorst:
There was a small bug in the initial code wherein the first line for each date was missing.

This fixes that issue and simplifies the logic of dividing into dates: when a new date is found, it makes sure that the first entries for `cc_band` and `iq_band` are defined so we can use `fillna` later to fill in the missing entries.

### DIFF
--- a/mock/environment/__init__.py
+++ b/mock/environment/__init__.py
@@ -2,7 +2,7 @@ import bz2
 import logging
 import os
 from datetime import timedelta
-from typing import Optional
+from typing import Optional, Union
 
 import astropy.units as u
 import numpy as np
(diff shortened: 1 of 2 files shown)


mock/environment/__init__.py
```diff
@@ -27,19 +27,19 @@ def _data_file_path(filename: str) -> str:
         return os.path.join('..', '..', 'data', filename)
```
stroncod, line 27: This return is alone?

Davidwk123 (approved):

stroncod (approved):

Davidwk123 (approved):

mock/environment/__init__.py
sraaphorst: Yeah, it's just a one-line method because we have to create four paths to files in the `data` path, so this just does it consistently.

PR #139, opened 2022-07-20
SCHED 163: Add barcode transformation to helpers

stroncod:


common/helpers/__init__.py
```diff
@@ -1,9 +1,11 @@
 from collections.abc import Iterable
+from lib2to3.pgen2.token import OP
```
sraaphorst, line 2: Just curious where you are using this (I don't see it anywhere) and if we need to rely on the `lib2to3` library for this.

stroncod: I don't know what is that or how that got there  😂 I'll remove it

sraaphorst: I was wondering what we were doing with using Python conversion code. 😂

sraaphorst (approved):

PR #141, opened 2022-07-21
Sched 179

Davidwk123:


mock/environment/__init__.py
```diff
@@ -167,6 +167,39 @@ def __init__(self):
...
+
+        if start_time.date() in self.site_data_by_night[site] and end_time.date() in self.site_data_by_night[site]:
+            start_time = round_seconds(start_time)
+            end_time = round_seconds(end_time)
+
+            if start_time <= end_time:
+                if start_time in set(self.site_data_by_night[site][start_time.date()][Env._time_stamp]) and end_time in set(self.site_data_by_night[site][end_time.date()][Env._time_stamp]):
+                    for night in self.site_data_by_night[site]:
+                        for index, data in self.site_data_by_night[site][night].iterrows():
+                            if data[Env._time_stamp] >= start_time and data[Env._time_stamp] <= end_time:
```
sraaphorst, line 191: This is fine but like I mentioned up there a few comments ago, you can make it a little easier to read by writing it as:
```python
if start_time <= data[Env._time_stamp] <= end_time:
```

mock/environment/__init__.py
```diff
@@ -167,6 +167,39 @@ def __init__(self):
...
+                    for night in self.site_data_by_night[site]:
+                        for index, data in self.site_data_by_night[site][night].iterrows():
+                            if data[Env._time_stamp] >= start_time and data[Env._time_stamp] <= end_time:        
+                                weather_list.append(data)
+                else:
+                    logging.info("Error: Starting/ending time is not stored! ")
+            else:
+                logging.info("Error: Starting date is past ending date!")
+        else:
+            logging.info("Error: Starting/ending date is not stored! ")
```
sraaphorst, line 198: Now we don't need any of these, because we don't care if the start / end date or time is not stored, and we handled the case of the starting date being past the ending date at the top of the function.

mock/environment/__init__.py
```diff
@@ -167,6 +167,39 @@ def __init__(self):
...
+
+        return weather_list
```
sraaphorst, line 200: Making these changes takes the runtime I was testing for a small dataset from 23 seconds down to 0.02 seconds, and we want this function to run fast.

So this is what going through your code and making the changes left it looking like for me:
```python
    def get_weather(self, site: Site, start_time: datetime, end_time: datetime) -> List[object]:
        """
        Returns list of weather data
        based off start and end times
        """
        if start_time > end_time:
            logging.warning(f'Weather request for invalid data range: {start_time} to {end_time}.')
            return []

        weather_list = []

        start_date = start_time.date()
        end_date = end_time.date()
        nights = [n for n in self.site_data_by_night[site] if start_date <= n <= end_date]

        for night in nights:
            for _, data in self.site_data_by_night[site][night].iterrows():
                if start_time <= data[Env._time_stamp] <= end_time:
                    weather_list.append(data)

        return weather_list
```

I precalculated `start_date` and `end_date` because otherwise, they might be recalculated in every iteration of the for comprehension:
```python
nights = [n for n in self.site_data_by_night[site] if start_date <= n <= end_date]
```
and that would increase the runtime.

sraaphorst (changes requested):
Good progress: we just need some adjusting for speed, and to allow for arbitrary weather queries instead of requiring the start time and the end time to specifically be in the data.

sraaphorst (approved):

PR #142, opened 2022-07-21
Sched 181: Ranker Modularization

stroncod:


components/selector/__init__.py
```diff
@@ -35,7 +35,7 @@ class Selector(SchedulerComponent):
     def select(self,
                sites: FrozenSet[Site] = ALL_SITES,
                night_indices: Optional[npt.NDArray[NightIndex]] = None,
-               ranker: Optional[Ranker] = None) -> Selection:
+               ranker: Optional[DefaultRanker] = None) -> Selection:
```
sraaphorst, line 38: This looks good but I'm a bit confused by this line here. Wouldn't you want this to still be `Ranker` so you can pass in other subclasses of `Ranker`?

stroncod: My bad!

sraaphorst (approved):
👍

PR #150, opened 2022-07-26
SCHED-192:Make Env service get_weather method return List[Variant]

Davidwk123:


sraaphorst (approved):
👍 This looks great to me.

services/environment/__init__.py
```diff
@@ -184,7 +186,26 @@ def get_weather(self, site: Site, start_time: datetime, end_time: datetime) -> L
...
+        pd.DataFrame(weather_list)
+
+        for weather in weather_list:
+            # extract the fields above
+            try:
+                variant = Variant(
+                    start_time = weather[Env._time_stamp],
+                    cc = CloudCover(weather[Env._cc_band]),
+                    iq = ImageQuality(weather[Env._iq_band]),
+                    wind_dir = weather[Env._wind_dir] * u.rad,
```
sraaphorst, line 198: Whoops... I kind of led you astray on this line. It should be:
```
wind_dir = Angle(...)
```
where `...` is what you already have there. Otherwise, AstroPy makes it a `Quantity` instead of an `Angle`, which will cause problems.

PR #154, opened 2022-08-01
SCHED-196: Modifies Process Manager

stroncod:
Adds Singleton class for all our singletons! (ProcessManager y PlanManager)

### DIFF
--- a/graphql_server/__init__.py
+++ b/graphql_server/__init__.py
@@ -1,6 +1,7 @@
 from copy import deepcopy
 from datetime import date, datetime
-from threading import Lock
+from tokenize import Single
+from common.meta import Singleton
 from typing import List, NoReturn
 import pytz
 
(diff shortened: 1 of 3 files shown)


graphql_server/__init__.py
```diff
@@ -1,6 +1,7 @@
 from copy import deepcopy
 from datetime import date, datetime
-from threading import Lock
+from tokenize import Single
```
sraaphorst, line 3: What does this do and where is it used? I don't see any reference to `Single` in the code.

graphql_server/__init__.py
```diff
@@ -85,22 +86,15 @@ def for_site(self, site: Site) -> 'SPlans':
         )
 
 
-class PlanManager:
+class PlanManager(metaclass=Singleton):
```
sraaphorst, line 89: Do we really need metaclasses? Can't we perhaps do something simpler?

https://realpython.com/python-metaclasses

graphql_server/__init__.py
stroncod: VS Code add some dumb stuff to autocomplete. Thanks, I'll delete it

graphql_server/__init__.py
stroncod: A Base class can get the same results but metaclass is the recommended singleton implementation: https://stackoverflow.com/questions/6760685/creating-a-singleton-in-python

sraaphorst: Okay, sounds good.

sraaphorst (approved):

graphql_server/__init__.py
sraaphorst: LOL happens to all of us sometimes.

PR #155, opened 2022-08-05
Sched 187: New folder structure and GraphQL integration

stroncod:
This adds 3 major changes:

- A new folder structure
- GraphQL integration with ProcessManger (Adds mutation for a new plans to be generated)
- Adds endpoint for subscribing to ODB changes (WIP to make it functional with a real front end)


app/graphql/schema.py
```diff
@@ -0,0 +1,46 @@
...
+
+@strawberry.type
+class Mutation:
+    @strawberry.mutation
+    async def new_schedule(self,
+                           new_schedule_input: CreateNewScheduleInput) -> bool:
+        try:
+            start, end = Time(new_schedule_input.start_time, format='iso', scale='utc'), Time(new_schedule_input.end_time, format='iso', scale='utc')
+        except ValueError:
+            raise ValueError("Invalid time format. Must be ISO8601.")
```
sraaphorst, line 27: Good stuff.

app/graphql/schema.py
```diff
@@ -0,0 +1,46 @@
...
+        manager = PM()
```
sraaphorst, line 28: Why did we change the name to `PM`? It seems really unintuitive to me.

app/mock/environment/__init__.py
```diff
@@ -8,7 +8,7 @@
 from astropy.time import Time
 from astropy.units import Quantity
 
-from common.minimodel import Site, Variant, CloudCover, ImageQuality
+from app.common.minimodel import Site, Variant, CloudCover, ImageQuality
 
 
 class Env:
```
sraaphorst, line 14: Here's a move I'm not sure about... why should `mock` be part of `app`?

app/plan_manager/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+    def set_plans(plans: List[Plans]) -> NoReturn:
+        """
+        Note that we are converting List[Plans] to List[SPlans].
+        """
+        PlanManager._lock.acquire()
+        calculated_plans = deepcopy(plans)
+        PlanManager._plans = [
+            SPlans.from_computed_plans(p) for p in calculated_plans
+        ]
+        PlanManager._lock.release()
```
sraaphorst, line 43: I think I like this better that the way it was before.

app/routers.py
```diff
@@ -0,0 +1,24 @@
...
+async def odb_changes_endpoint():
+    s = Session(url=config.graphql.url)
+    while True:
+        resp = await s.subscribe_all()
+        if resp:
+            print(resp)
+            # Run new Schedule
+            mode = TaskType.STANDARD
+            scheduler = build_scheduler()
+            PM().add_task(datetime.now(), scheduler, mode)
```
sraaphorst, line 22: This syntax just seems confusing to me... I assume `PM` is a singleton... it seems like we should be getting an instance of `PM` rather than using the create new object syntax.

app/scheduler/scheduler.py
```diff
@@ -6,30 +6,35 @@
 from astropy.time import Time, TimeDelta
```
sraaphorst, line 6: Kind of weird to have this doubly nested package name, no?

main.py
```diff
@@ -0,0 +1,42 @@
...
+    asyncio.run(new_schedule(manager, scheduler))
+    return JSONResponse(status_code=200,
+                        content={ "message": f"New schedule created: {plan_manager.get_plans()}"})
+'''
+
+
+if __name__ == "__main__":
+    manager = ProcessManager(size=config.process_manager.size,
+                             timeout=config.process_manager.timeout)
+    uvicorn.run(app, host='127.0.0.1', port=8000)
```
sraaphorst, line 42: Good. This was a change I wanted to make but David found this would work... it was not in the uvicorn API as far as I could tell.

main.py
```diff
@@ -0,0 +1,42 @@
...
+'''
+@app.get("/new", include_in_schema=False)
+def new() -> JSONResponse:
+    start = Time("2018-10-01 08:00:00", format='iso', scale='utc')
+    end = Time("2018-10-03 08:00:00", format='iso', scale='utc')
+    scheduler = Scheduler(config, start, end, plan_manager)
+    asyncio.run(new_schedule(manager, scheduler))
+    return JSONResponse(status_code=200,
+                        content={ "message": f"New schedule created: {plan_manager.get_plans()}"})
+'''
```
sraaphorst, line 36: Why commented out? Will this still work?

scripts/run_scheduler.py
```diff
@@ -161,9 +161,9 @@
     print_plans(plans)
 
     if run_graphql_server:
-        import graphql_server
-        plan_manager = graphql_server.PlanManager()
+        import graphql
+        plan_manager = graphql.PlanManager()
         plan_manager.set_plans(plans)
-        graphql_server.start_graphql_server()
+        graphql.start_graphql_server()
```
sraaphorst, line 167: This looks good.

sraaphorst (commented):
The:
```
app/data/gn_wfs....
app/data/gs_wfs....
```
files are not associated with the app, but with David's `Env` service, so they should be moved into a data directory in there.

While I think there's a lot that does improve the code here, I'd just like to know about some of these design decisions and why you made them. For example, why did you move `common` into `app/common`, especially since elements of `common` are used outside of `app`?

Also, I think we should finally get rid of `old` completely. Tell Bryan to grab what he needs from here (i.e. just copy the whole structure over if he needs for GreedyMax) and then delete it from the project.

Overall, some insight would be good. The GraphQL organization is much better, but there are some things I have concerns about.

app/scheduler/scheduler.py
stroncod: Yes, but this should be scrapped and change instead from a Builder

app/graphql/schema.py
stroncod: Copilot suggestion. It's changed now.

app/mock/environment/__init__.py
stroncod: Mock at the end should be removed in favor of `services` right?

app/routers.py
stroncod: `()` should return the instance based on the metaclass. It removes the need for the `instance()` method.

main.py
stroncod: Removed

stroncod:
Some modifications were added to some suggestions. More discussion is pending.

app/graphql/schema.py
sraaphorst: 👍

app/mock/environment/__init__.py
sraaphorst: I think so... I don't think we'll need to keep `mock` if we have `services` that are actually implemented.

app/routers.py
sraaphorst: Yeah, that's what I figured. It just seems a bit confusing to me since most people would assume you're creating a new `PM` but we can leave it.

main.py
sraaphorst: 👍

sraaphorst:
I think I really like this reorganization as it is now and it's something that needed to be done, so thanks for taking care of it!
I'm just going to look it over a bit more, but I think we can merge this soon as it is.

sraaphorst (approved):
Alright, let's merge this and see how it looks.

PR #158, opened 2022-08-18
SCHED 202: Working with lucupy external package 

stroncod:
The lucupy package replace the old common folder. 
This PR is to discuss if and how is going to be implemented.

PR #159, opened 2022-08-25
Sched 205: Subscriber 

stroncod:
Based on PR #158. 

New `Subscriber` that separate all the subscription logic and creates his own service.


sraaphorst (approved):
Looks good. I'll have to see everything independently to get a sense of it.

PR #162, opened 2022-08-31
SCHED-177: License, copyright, code cleanup

sraaphorst:
This adds:
* the `LICENSE`;
* the copyright header to each `.py` file; and
* a script to check if any `.py` files are missing the copyright header.

It also:
* removes a bunch of unnecessary files from data;
* cleans up the headers and optimizes them;
* reformats the code so that it matches Python best practices;
* makes minor changes where they were come across.


scripts/check_headers.py
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
```
stroncod, line 1: Awesome!

stroncod (approved):
Thanks for the clean up!

PR #167, opened 2022-09-07
SCHED-217: Remove Subscriber files

stroncod:


sraaphorst (approved):
Not quite sure why these are being removed, but why not?

stroncod:
> Not quite sure why these are being removed, but why not?

It has it's own repo now: https://github.com/gemini-hlsw/scheduler-subscriber

PR #174, opened 2022-10-11
SCHED-168 and SCHED-226: ResourceMock rewrite and lookup tables for FPU barcodes

sraaphorst:
Not quite ready for merge, but request for comments.
Added back data files.


stroncod (approved):

sraaphorst:
The error should be fixed now, but I want to move some things to `OcsProgramProvider.`

stroncod (approved):
I'm confused on the addition of Site in some methods but looks good.

sraaphorst:
The site was added because we're using barcodes as the `Resource` IDs, and we want to differentiate between sites since some of the barcodes might overlap.

sraaphorst:
For example, `IFU-R` has barcodes 1000009 or something like that at both sites, but might not be active at both sites.

PR #176, opened 2022-10-12
Bugfix/parabolic metric

bryanmiller:
Fixed some bugs in the metric calculation
- used list instead of dictionary since order is important
- use the correct parameter for the exponent in the metric
- cleanup


bryanmiller:
No, I don't think we need comp_exp.

PR #178, opened 2022-10-20
SCHED-240: Major rewrite of ResourceMock and test cases

sraaphorst:
1. Cleaned up and added functionality to ResourceMock.
2. Fixed some bugs: might interfere with https://github.com/gemini-hlsw/scheduler/pull/177
3. Added test cases for ResourceMock.


app/core/output/__init__.py
```diff
@@ -27,7 +26,7 @@ def print_program_from_provider(filename=os.path.join('data', 'GN-2018B-Q-101.js
         data = json.loads(f.read())
 
     program = provider.parse_program(data['PROGRAM_BASIC'])
-    print_program(program)
+    program.show()
```
stroncod, line 29: Annoying changes, am I right?

mock/resource/__init__.py
```diff
@@ -16,91 +16,126 @@
 from definitions import ROOT_DIR
 
 
-# TODO: Should this be a Singleton?
+@final
```
stroncod, line 19: Did not know about `final`. Thanks!

stroncod (approved):

app/core/components/collector/test/test_proper_motion.py
```diff
@@ -15,7 +15,7 @@ def time():
 
 @pytest.fixture
 def target():
-    return SiderealTarget('', set(), TargetType.OTHER,
+    return SiderealTarget('', frozenset(), TargetType.OTHER,
```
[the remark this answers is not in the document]
sraaphorst: We should be using `frozenset` whenever possible, especially for something like magnitudes because they are not going to change. Is there a reason why you see that we should use `set` over `frozenset`?

These objects should be as immutable as possible, IMO, unless we have a very good reason for them to not be.

app/core/output/__init__.py
sraaphorst: Annoying how? I think you moving `show` into `Program` is fine.

mock/resource/__init__.py
sraaphorst: I knew about `Final` in declaring constants, but not in classes and methods until just a few days ago.

sraaphorst: (as a decorator)

app/core/components/collector/test/test_proper_motion.py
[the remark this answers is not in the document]
stroncod: I agree but frozentset was causing issues I recall? or I'm just getting confused with some other part of the code that we do `frozenset` to `set`?

app/core/output/__init__.py
stroncod: lol just a joke because was my change

stroncod:
Let's merge it! :D

PR #179, opened 2022-10-24
SCHED-232: Change Collector and Selector to use ResourceMock

sraaphorst:
Before, the `Collector` just returned a test static set of `Resource` that matched the observations passed in.
Now, the `Collector` uses the `ResourceMock` to actually return the viable `Resource` for a `Site` for a given range of `NightIndex`.

Furthermore, beforehand, this information was calculated for every group. Now, instead, the `Resource` availability per active `Site` per `NightIndex` is precalculated in the `Selector` and passed on as necessary to avoid recalculation.

In addition, this PR includes a couple small independent tasts:
1. Cleanup of `DBManager` (removal of some unused imports, spacing).
2. Removal of `plan.db` and `plan.lock` created by `DBManager` using the `shelve` library since this should not be in GitHub. These were also added to `.gitignore`.


app/core/components/collector/__init__.py
```diff
@@ -434,55 +436,12 @@ def load_programs(self, program_provider: ProgramProvider, data: Iterable[dict])
         if bad_program_count:
             logging.error(f'Could not parse {bad_program_count} programs.')
 
-    @staticmethod
-    def available_resources(site: Site,
-                            night_idx: NightIndex) -> FrozenSet[Resource]:
+    def available_resources(self,
+                            site: Site,
+                            night_indices: npt.NDArray[NightIndex]) -> List[FrozenSet[Resource]]:
```
stroncod, line 441: This could be not too efficient if we want to look for a specific value right. O(n^2) search, I think we are cover now because we just need to check that a value is in there.

stroncod (approved):
Good stuff!

PR #181, opened 2022-10-26
SCHED-244: Reset Observations

sraaphorst:
This PR resets `Observation` data. This includes:
1. Setting the `Observations` with specified `ObservationStatus` to having `ObservationStatus.READY`.
2. Traversing the sequence (`Atom`s) of all `Observation`s and clearing the recorded time used.

This method was made static so it can be called from anywhere given an `Iterable` of `Program`s, and allows optional additional filtering on `Program`s and `Observation`s.

Some additional very minor changes included with this PR.


app/core/components/collector/test/test_clear_observation_info.py
```diff
@@ -0,0 +1,34 @@
...
+import os
+
+from lucupy.minimodel.observation import ObservationStatus
+
+from app.core.components.collector import Collector
+from app.core.programprovider.ocs import read_ocs_zipfile, OcsProgramProvider
+from definitions import ROOT_DIR
+
+
+def test_clear_observations():
```
stroncod, line 14: Great thing we keep adding test! Coverage is happier now

stroncod (approved):

PR #184, opened 2022-10-31
Cleanup: Getting rid of warnings, optimizing imports, and adding missing headers.

sraaphorst:


app/core/scheduler/modes.py
```diff
@@ -62,10 +66,11 @@ def schedule(self, start: Time, end: Time):
         # Save to database
         PlanManager.set_plans(plans)
 
+
 class OperationMode(SchedulerMode):
+    def schedule(self, start: Time, end: Time):
```
sraaphorst, line 71: @stroncod Can you check this to make sure it's right? I was getting complaints about function overloads that did not match the original signature.

app/core/scheduler/modes.py
```diff
@@ -1,35 +1,39 @@
...
     def schedule(self, start: Time, end: Time):
         pass
-    
+
     def __str__(self) -> str:
         return self.__class__.__name__
 
+
 class SimulationMode(SchedulerMode):
+    def schedule(self, start: Time, end: Time):
```
sraaphorst, line 32: Same with this one.

app/core/scheduler/modes.py
stroncod: That's odd because I'm using the same declaration as above in the abstract class, so there shouldn't be any overload. if you can drop me the specific error in Slack to take a look would be great. Thanks!

stroncod (approved):
I'll check the overload issue. Let's merge the rest.

app/core/scheduler/modes.py
sraaphorst: Sure. The superclass has:

```python
class SchedulerMode(ABC):
    @abstractmethod
    def sechedule(sefl, start: Time, end: Time):
        ...
```

and the override was:
```python
class SimulationMode(SchedulerMode):
    def schedule(self):
        ....
```

so you can see that `start: Time` and `end: Time` were missing from the override.

I was getting this:

```
Signature of method 'SimulationMode.schedule()' does not match signature of the base method in class 'SchedulerMode'
```

I won't merge this one until we've had a chance to discuss it tomorrow morning.

(Same with the other comment here.)

stroncod: Ok! I checked and is just a mistake when I wrote the method as is not really implemented yet.

PR #188, opened 2022-11-21
Sched 257: Add sources to config 

stroncod:
This is based on #187.

### DIFF
--- a/app/core/builder/blueprint.py
+++ b/app/core/builder/blueprint.py
@@ -1,6 +1,6 @@
 # Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
 # For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
-from typing import FrozenSet, Union, List, Iterable
+from typing import FrozenSet, Union, List, Any
 from enum import Enum
 
 from astropy.time import TimeDelta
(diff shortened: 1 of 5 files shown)


app/core/builder/blueprint.py
```diff
@@ -12,6 +13,28 @@
 
 from app.config import config, ConfigurationError
 from app.core.components.optimizer.dummy import DummyOptimizer
+from mock.resource import ResourceMock
+from mock.environment import Env
```
sraaphorst, line 17: Unsure why we need these here? Isn't the point to load the configs we need? We're going to throw away the Env mock, at least.

app/core/builder/blueprint.py
```diff
@@ -171,17 +131,38 @@ def __init__(self, algorithm: str) -> None:
...
     def __iter__(self):
         return iter((self.algorithm))
 
+class SourcesBlueprint(Blueprint):
+    
+    class ResourceSources(Enum):
+        MOCK = ResourceMock()
+        # TODO: As in full fledge service? I'm not sure about this name 
+        # so suggestions are welcome. 
+        FULL = None
```
sraaphorst, line 147: Maybe `GPP`?

sraaphorst:
Merging despite some small concerns.

app/core/builder/blueprint.py
stroncod: Same as the other options blueprints serve a way to change sources in runtime. Instead of using the default config and have to restart the service

app/core/builder/blueprint.py
```diff
@@ -171,17 +131,38 @@ def __init__(self, algorithm: str) -> None:
     def _parse_optimizer(algorithm_name: str):
         # TODO: Enums are needed but for now is just Dummy
         # TODO: When GMax is ready we can expand
-        if algorithm_name in 'Dummy':
+        if algorithm_name in 'DUMMY':
```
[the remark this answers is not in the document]
stroncod: You are right. It should be:
```
algorithm_name is Optimizers.DUMMY
```

app/core/builder/blueprint.py
stroncod: Sounds way better, I agree!

sraaphorst:
This seems like it might be breaking some things, so I am going to try reverting it.

PR #190, opened 2022-12-02
Sched 252 #1

bryanmiller:
An initial GM optimizer. This is based on the dummy optimizer and does not yet handle groups, atoms, or use the visibilities for placing observation. It fixes a few bugs in the dummy optimizer. It does not split, to fill in a night it iterates over observations in order of decreasing score until it finds something to schedule in the remaining time.


bryanmiller:
need to put app/core/components/collector/__init__.py back

PR #194, opened 2022-12-02
GreedyMax refactoring - step 2

bryanmiller:
This is based on the dummy optimizer and does not yet handle groups, atoms, or use the visibilities for placing observation. It fixes a few bugs in the dummy optimizer. It does not split, to fill in a night it iterates over observations in order of decreasing score until it finds something to schedule in the remaining time.

### DIFF
--- a/app/core/components/optimizer/greedymax.py
+++ b/app/core/components/optimizer/greedymax.py
@@ -71,8 +71,6 @@ class GreedyMaxOptimizer(BaseOptimizer):
             # else:
             #     print('group not added')
 
-            max_score = 0.0         # maximum score in time interval
-            max_group = {}
             maxscores = []
             groups = []
             ids = [] # group index for the scores
@@ -89,6 +87,8 @@ class GreedyMaxOptimizer(BaseOptimizer):
                     # print(group.group.id, group.group.exec_time(), smax)
 
             # sort
+            max_score = 0.0  # maximum score in time interval
+            max_group = {}
             jj = np.flip(np.argsort(maxscores))
             for ii in range(len(jj)):
                 max_score = maxscores[jj[ii]]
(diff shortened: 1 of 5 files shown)


app/core/components/ranker/default.py
```diff
@@ -194,6 +194,12 @@ def _score_obs(self, program: Program, obs: Observation) -> Scores:
                                               np.ones(1, dtype=int) * program.band,
                                               np.ones(1) * 0.8,
                                               program.thesis)
+        print(f'{obs.id:20}  exec: {obs.exec_time().total_seconds() / 3600.:.2f} '\
```
stroncod, line 197: This prints should either be in `logging` or removed as is not a good practice to have direct prints in the code.

stroncod (approved):

stroncod:
There are some minor changes but we could go ahead and merge it.

app/core/components/optimizer/greedymax.py
```diff
@@ -1,10 +1,139 @@
...
+
+
+    def setup(self, program_info: Mapping[ProgramID, ProgramInfo]) -> GreedyMaxOptimizer:
+        """
+        Preparation for the optimizer e.g. create chromosomes, etc.
+        """
+        self.groups = []
+        # self.scores = []
+        for p in program_info.values():
+            self.groups.extend([g for g in p.group_data.values() if g.group.is_observation_group()])
```
sraaphorst, line 50: Do we want to change this now to use the mapping of `UniqueGroupID` to `GroupData` now that it's available?

app/core/components/optimizer/greedymax.py
```diff
@@ -1,10 +1,139 @@
...
+            # Suggestion from Seb
+            # self.groups.extend([g for g in p.group_data.values() if all(x.is_observation_group() for x in g.group.children)])
+            # self.scores.extend([s for s in p.group_data])
+        return self
+
+    def _run(self, plans: Plans):
+
+        print()
```
sraaphorst, line 58: I agree with Sergio that we shouldn't be including `print` statements in PRs that we are going to merge, and should instead be logging.

sraaphorst (approved):
I'm concerned about the `print` statements being added to the `DefaultRanker`... if they were confined to GreedyMax or were just used for testing prior to merging and then removed, I would feel better, but I understand that we're still tinkering with the scoring a bit so I'll hesitantly approve.

PR #208, opened 2022-12-21
SCHED-282: Moved to using ROOT_GROUP_ID.

sraaphorst:


app/core/programprovider/ocs/test/test_ocs_api.py
```diff
@@ -545,12 +545,11 @@ def create_minimodel_program() -> Program:
...
     # *** ROOT GROUP ***
     root_children = [sched_group, gnirs1_group, gmosn1_group]
-    # root_children = [sched_group, gmosn1_group, gnirs1_group]
 
     root_group = AndGroup(
-        id='root',
+        id=ROOT_GROUP_ID,
         program_id=program_id,
-        group_name='root',
+        group_name=ROOT_GROUP_ID,
```
stroncod, line 552: if group_name and id are the same, what's the point of having group_name? Maybe that attribute should be Optional.

stroncod (approved):

app/core/programprovider/ocs/test/test_ocs_api.py
sraaphorst: They won't usually be the same. The group names are often human readable things like:
```
WBIS_2427+2830 - [1] GNIRS Spectroscopy 0.15"/pix 32 l/mm grating SXD 1.0 arcsec wavelength < 2.5
```
I'm just making them equal here because in the case of the root group, they are the same, but that's the only case.

In the case of the root group, technically, they could be different, but since the root group is a Scheduler-only concept, they probably wouldn't ever be. It might make sense to make the `group_name` optional based on this and just set for `None` for the `root_group`. Something to think about that won't affect anything outside of the Scheduler, I guess.

PR #209, opened 2022-12-22
SCHED-270: deepcopy of structures used in Selection

sraaphorst:
This is the partner PR of SCHED-270 in the lucupy project:
https://github.com/gemini-hlsw/lucupy/pull/27

It performs deep copies of structures returned in the `Selection` so that they can be modified for time accounting and will not alter the original data structures.

### DIFF
--- a/app/core/calculations/groupinfo.py
+++ b/app/core/calculations/groupinfo.py
@@ -40,8 +40,11 @@ class GroupInfo:
 class GroupData:
     """
     Associates Groups with their GroupInfo.
+    group is a deep copy.
     """
+    # Deep-copied.
     group: Group
+
     group_info: GroupInfo
 
     def __iter__(self):
--- a/app/core/calculations/groupinfo.py
+++ b/app/core/calculations/groupinfo.py
@@ -4,12 +4,14 @@
 from dataclasses import dataclass
 from typing import Dict, List
 
-import numpy.typing as npt
+from lucupy.decorators import immutable
 from lucupy.minimodel import Conditions, Group, GroupID
+import numpy.typing as npt
 
 from .scores import Scores
 
 
+@immutable
 @dataclass(frozen=True)
 class GroupInfo:
     """
@@ -36,6 +38,7 @@ class GroupInfo:
     scores: Scores
 
 
+# Leave this non-immutable as group might change, and we want that change to perpetuate.
 @dataclass(frozen=True)
 class GroupData:
     """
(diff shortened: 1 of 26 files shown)


app/core/calculations/groupinfo.py
```diff
@@ -40,8 +40,11 @@ class GroupInfo:
 class GroupData:
     """
     Associates Groups with their GroupInfo.
+    group is a deep copy.
     """
+    # Deep-copied.
```
stroncod, line 45: This might be an overreach but would this increment our RAM consumption ? Probably nothing to worry about, just a thought.

app/core/calculations/nightevents.py
```diff
@@ -35,8 +36,31 @@ class NightEvents:
     moonset: Time
 
     # Information for Julian years.
-    _JULIAN_BASIS: ClassVar[float] = 2451545.0
-    _JULIAN_YEAR_LENGTH: ClassVar[float] = 365.25
+    _JULIAN_BASIS: ClassVar[float] = field(default=2451545.0, init=False, repr=False, compare=False)
```
stroncod, line 39: I think this variables are in some form in lucupy.sky

services/horizons/__init__.py
```diff
@@ -16,12 +16,13 @@
...
 
 
+@final
 @dataclass
 class HorizonsAngle:
     """
     Angle in radians.
     """
-    microarcsecsPerDegree: float = 60 * 60 * 1000 * 1000
+    MicroarcsecsPerDegree: Final[float] = 60 * 60 * 1000 * 1000
```
stroncod, line 25: Even better `MICRO_ARCSECS_PER_DEGREE` I think is python guideline for variables like that. (globals at least)

stroncod (approved):

services/horizons/__init__.py
sraaphorst: Makes sense, because we also have constants written in CAPS in lots of other places. I'll change it.

app/core/calculations/nightevents.py
sraaphorst: Thanks! I'll check and remove them if they are.

app/core/calculations/groupinfo.py
sraaphorst: Doubtful... at first I had thought that we might be able to get away with copying less, but I was wrong, so we probably could have done a straightforward deepcopy.

Nonetheless, I learned a few things... so... 😅

PR #210, opened 2022-12-26
SCHED-285: Add CORS Middleware to FastAPI

stroncod:
This blocks outside request to the Scheduler and satisfies the FrontEnd server requirements.


sraaphorst (approved):
👍

PR #211, opened 2022-12-27
SCHED-260: Fix Docker-image building and re-structuring of the project

stroncod:
This PR looks big but is not, the only true change is on the Dockerfile. The rest is whole chunk of the code moved to create a new structure.  This allow us to copy just two files in the Dockerfile making the building easier. 

WIP. DO NOT MERGE


sraaphorst:
It looks like this is breaking a bunch of tests?

stroncod:
Important changes:

- change the main folder to the name of the project (scheduler), same as in lucupy. 
- As an effect to the change above, the `Scheduler` class is now called `Service`
- To avoid circular dependency on testing. the service module `__init__.py` file does not have the imports for the sub-modules inside. This require to import directly for the sub-modules.
- `graphql` module is change to `grapqhl_mid` to not have issues with `graphql-core` that is imported in the same way.
- test are now in there own folder (and not being copied to the Dockerfile). This follows lucupy structure as well.

scheduler/core/service/__init__.py
```diff
@@ -1,4 +1,4 @@
 # Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
 # For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
-from .scheduler import build_scheduler, Scheduler
-from .modes import SchedulerModes, OperationMode, SimulationMode, ValidationMode
+# from .service import build_scheduler, Service
+# from .modes import SchedulerModes, OperationMode, SimulationMode, ValidationMode
```
sraaphorst, line 4: Why commented out instead of just removed?

scheduler/db/dbmanager.py
```diff
@@ -6,7 +6,7 @@
 from fcntl import flock, LOCK_SH, LOCK_EX, LOCK_UN
 from typing import List
 
-from app.graphql.scalars import SPlans
+from scheduler.graphql_mid.scalars import SPlans
```
sraaphorst, line 9: What does the `_mid` mean?

sraaphorst (approved):
I have a couple questions but this looks fine.

scheduler/db/dbmanager.py
stroncod: middleware

scheduler/core/service/__init__.py
stroncod: Because we might want to go back to that if we change the code to remove the circular dependency.

PR #212, opened 2023-01-24
SCHED-288: Updated run_scheduler.py to use new techniques to build Scheduler components.

sraaphorst:


stroncod (approved):
Thanks for the fix!

PR #214, opened 2023-02-14
SCHED 297: GraphQL error handling with shelve 

stroncod:
There were some errors when the DB was not initialize but it was query anyway, making the instance in Heroku useless, causing problems on the UI as well. This fix these and add testing to our endpoints so we can check if any changes break them.

### DIFF
--- a/scheduler/db/dbmanager.py
+++ b/scheduler/db/dbmanager.py
@@ -4,7 +4,7 @@
 import shelve
 from contextlib import contextmanager
 from fcntl import flock, LOCK_SH, LOCK_EX, LOCK_UN
-from typing import List
+from typing import List, NoReturn
 
 from scheduler.graphql_mid.scalars import SPlans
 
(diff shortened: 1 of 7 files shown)


scheduler/db/dbmanager.py
```diff
@@ -23,12 +23,14 @@ class DBManager:
     def __init__(self, db_path):
         self.db_path = db_path
 
-    def read(self):
+    def read(self) :
```
sraaphorst, line 26: Just curious: why the space before the `:` and no return type?

scheduler/db/planmanager.py
```diff
@@ -28,15 +28,21 @@ def get_plans() -> List[SPlans]:
         This is to ensure that the plans are not corrupted after the
         lock is released.
         """
-        plans = deepcopy(db.read())
-        return plans
+        try:
+            plans = deepcopy(db.read())
+            return plans
+        except Exception as err:
```
sraaphorst, line 34: Can we be more specific as to the types of `Exception`s? PEP-8 says we should avoid using naked `except` or catch-all `except`. Is it possible that this could raise an `Error` which I don't think would be caught by an `except Exception`?

tests/unit/scheduler/graphql_mid/test_graphql_mutations.py
```diff
@@ -0,0 +1,28 @@
+import pytest
```
sraaphorst, line 1: We should have the header here as well as in:
`tests/unit/scheduler/graphql_mid/__init__.py`

tests/unit/scheduler/graphql_mid/test_graphql_mutations.py
```diff
@@ -0,0 +1,28 @@
...
+from scheduler.graphql_mid.server import  schema
```
sraaphorst, line 2: Double space after import.

tests/unit/scheduler/graphql_mid/test_graphql_queries.py
```diff
@@ -0,0 +1,30 @@
+from scheduler.graphql_mid.server import  schema
```
sraaphorst, line 1: Header missing here.

sraaphorst (approved):
Just a few small change requests and the exception note: this looks good.

PR #216, opened 2023-02-17
SCHED-304: Resource service update and addition of filters

sraaphorst:
This PR is to account for the changes to the OCS resource service.

Due to the substantial changes to the Excel telescope schedule / configuration sheet and requirement changes and clarifications, enormous changes have been made to the `OcsResourceService`, which used to just return a `FrozenSet[Resource]` for a given `Site` and `date`.

With the changes to the telescope configuration, all of the following data had to be represented:
1. PV nights
2. Classical nights
3. Instrument nights
4. Program priority nights
5. Partner nights
6. ToO allowances / disallowances
7. LGS configuration

The following Google drive folder contains documents relevant to this PR:
https://drive.google.com/drive/u/0/folders/1eQgohFrbYUkmPXuxdUpN3pv_vzMoMwN8

These include:
1. The telescope configuration / schedule Excel spreadsheet.
1. A standardization guide for the spreadsheet, with clarification of the requirements for the different possible modes.
2. A mini design document detailing the implementation of `OcsResourceService`.

Note that `OcsResourceService` now attempts to pull the Excel spreadsheet from this location so that it has the most recently update copy. Any changes to the Excel spreadsheet should follow the standardization guide.


scheduler/services/resource/test/__init__.py
```diff
@@ -1,4 +1,4 @@
 # Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
 # For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
 
-# This file is needed for the test_resource_mock.py file to run.
+# This file is needed for the test_ocs_resource_service.py file to run.
```
stroncod, line 4: Test are now located in the tests folder at the top.

scheduler/services/resource/ocs_resource_service.py
```diff
@@ -0,0 +1,568 @@
...
+    It can then be queried to receive a set of Resource (usually with barcode IDs, except for instruments) for a
+    given site on a given night.
+
+    It caches and reuses Resources by ID as best as possible to minimize the number of Resource objects existing
+    at any given time. Since Resource is immutable, this should be fine.
+
+    Note that this is a Singleton class, so new instances do not need to be created.
+    """
+    # The Google ID of the telescope configuration file.
+    _SITE_CONFIG_GOOGLE_ID: Final[str] = '1QRalQNEaX-bcyrPG6mfKnv01JVMaGHwy'
```
stroncod, line 41: There is not issue with this being public?

sraaphorst: I don't think so. We had the actual spreadsheet itself checked into the repo before... now we just have a link to it instead on the Google drive. I was thinking at first that it should be perhaps kept a secret, but I doubt old telescope configs are secret. We can confirm with Bryan and Kristin on Tuesday.

scheduler/services/resource/test/__init__.py
sraaphorst: Thanks! I'll move it.

scheduler/services/resource/__init__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
```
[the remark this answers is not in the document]
sraaphorst: Yes. *slaps head*
I completely forgot about this. It definitely makes sense as a script and it needs a bit of updating.

scheduler/core/builder/blueprint.py
```diff
@@ -13,8 +13,8 @@
 
 from scheduler.config import config, ConfigurationError
 from scheduler.core.components.optimizer.dummy import DummyOptimizer
-from scheduler.external.resource import ResourceMock
-from scheduler.external.environment import Env
+from scheduler.services.resource import OcsResourceService
```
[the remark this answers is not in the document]
sraaphorst: I don't really see any way around it, given that the information will be coming from completely different sources. Perhaps we can put a layer between `OcsResourceService` and `GppResourceService`, but right now, since I don't think there is a `GppResourceService`, it makes much sense to think about that now.

Something that I thought about that could get very confusing: if someone requests plan generation that begins before OCS becomes inactive and ends when GPP becomes active.

scheduler/services/resource/test/__init__.py
sraaphorst: Never mind... since we don't have any official tests, I just removed it.

scheduler/core/builder/blueprint.py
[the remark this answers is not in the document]
sraaphorst: In fact, since the mock is gone, we can probably remove this completely from the builder / blueprint / config.

scheduler/services/resource/ocs_resource_service.py
stroncod: Great, if is per document than should be not issue.

scheduler/core/builder/blueprint.py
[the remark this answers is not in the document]
stroncod: > I don't really see any way around it, given that the information will be coming from completely different sources. Perhaps we can put a layer between `OcsResourceService` and `GppResourceService`, but right now, since I don't think there is a `GppResourceService`, it makes much sense to think about that now.
> 
> Something that I thought about that could get very confusing: if someone requests plan generation that begins before OCS becomes inactive and ends when GPP becomes active.

Yes I think this might be an issue for those special cases.

stroncod (approved):
Issues fixed!

PR #218, opened 2023-02-23
SCHED-298: Modify mutation to accept site as a parameter

stroncod:


scheduler/core/service/service.py
```diff
@@ -39,14 +42,16 @@ def __call__(self):
 
 
 def build_scheduler(start: Time = Time("2018-10-01 08:00:00", format='iso', scale='utc'),
-                    end: Time = Time("2018-10-03 08:00:00", format='iso', scale='utc')) -> Service:
+                    end: Time = Time("2018-10-03 08:00:00", format='iso', scale='utc'),
+                    sites: FrozenSet[Site] = ALL_SITES) -> Service:
```
sraaphorst, line 46: LOL I just picked these dates randomly but they look like they'll live on for awhile in infamy.

scheduler/graphql_mid/schema.py
```diff
@@ -23,15 +24,19 @@ async def new_schedule(self,
...
             return NewScheduleError(error='Invalid time format. Must be ISO8601.')
         else:
-            pm = setup_manager()
-            pm.add_task(datetime.now(), scheduler, TaskType.STANDARD)
-            # await asyncio.sleep(10) # Use if you want to wait for the scheduler to finish
-            return NewScheduleSuccess(success=True)
+            try:
+                pm = setup_manager()
+                pm.add_task(datetime.now(), scheduler, TaskType.STANDARD)
+                await asyncio.sleep(10) # Use if you want to wait for the scheduler to finish
```
sraaphorst, line 35: Seems like kind of an arbitrary number? Why 10 s?

sraaphorst (approved):
I had one question but otherwise, this looks good. 👍

scheduler/graphql_mid/schema.py
stroncod: It is. Is a hack to force the waiting so I can test that in the UI. It should be removed in further changes

PR #219, opened 2023-03-01
Gmax refactor

bryanmiller:
I rebased my branch for the GreedyMax refactoring. The refactoring is still in early stages.
Also, this fixes inconsistencies between build_collector and CollectorBlueprint so that run_scheduler.py will run.


bryanmiller:
I have more changes so I'll submit a new PR later.

PR #224, opened 2023-03-07
Gmax refactor

bryanmiller:
Greedy-max changes due to rebase.

### DIFF
--- a/scheduler/core/components/optimizer/timeline.py
+++ b/scheduler/core/components/optimizer/timeline.py
@@ -1,0 +1,62 @@
+EMPTY = -1
+UNSCHEDULABLE = -2
+
+from dataclasses import dataclass
+from typing import List, NoReturn, Mapping, Optional, Sized, Tuple
+from datetime import datetime, timedelta
+
+from lucupy.minimodel import Observation, ObservationID, Site
+from app.core.calculations.nightevents import NightEvents
+
+import numpy as np
+
+
+@dataclass
+class Timeline:
+    """
+    Nightly plan for a specific Site for the GreedyMax optimizer. Each plan is a timeline array with one
+    entry for each time slot for the night. Each value needs to be a pointer or index to the observation
+    scheduled in that slot.
+    """
+    start: datetime
+    end: datetime
+    time_slot_length: timedelta
+    site: Site
+    _total_time_slots: int
+
+    def __post_init__(self):
+        self.schedule = np.full(self._total_time_slots, EMPTY)
+        self.is_full = False
+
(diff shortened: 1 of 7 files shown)


scheduler/core/components/optimizer/timeline.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def __init__(self, night_events: Mapping[Site, NightEvents], night_idx: int):
+
+        self.timelines = {}
+        self.night = night_idx
+        for site, ne in night_events.items():
+            if ne is not None:
+                self.timelines[site] = Timeline(ne.local_times[night_idx][0],
+                                        ne.local_times[night_idx][-1],
+                                        ne.time_slot_length.to_datetime(),
```
sraaphorst, line 48: Interesting... I didn't know that an AstroPy `TimeDelta` had a `to_datetime()` method that converted the object into a `datetime.timedelta` Python object.

Since we only have a single entry for `time_slot_length`, there isn't much benefit to using a `TimeDelta` for it, I don't think. I thought it would play nicer with AstroPy `Time` objects, but it seems you can add either an AstroPy `TimeDelta` or a regular old `datetime.timedelta` to an AstroPy `Time` object and both work just as you would expect.

I won't change this now, but I'll make a Jira task to look into it.

scheduler/core/components/optimizer/timeline.py
```diff
@@ -0,0 +1,62 @@
...
+    time_slot_length: timedelta
+    site: Site
+    _total_time_slots: int
+
+    def __post_init__(self):
+        self.schedule = np.full(self._total_time_slots, EMPTY)
+        self.is_full = False
+
+    def has(self, obs: Observation) -> bool:
+        return any(visit.obs_id == obs.id for visit in self.visits)
```
sraaphorst, line 32: A proposal here: instead of using a `has` method, I suggest we use the `__contains__` dunder method, i.e.:

```python
def __contains__(self, obs: Observation) -> bool:
    return any(visit.obs_id == obs.id for visit in self.visits)
```

Then instead of calling:
```python
timeline.has(obs)
```
or:
```python
not timeline.has(obs)
```

We can call instead:
```python
obs in timeline
```
or:
```python
obs not in timeline
```

sraaphorst (approved):
I'm going to approve this, but I do recommend we change `Timeline.has` to `Timeline.__contains__` for best Python practices. If you could do that and push another commit, it won't change the approval status, and you can still merge without @stroncod or I having to re-approve.

sraaphorst:
(BTW, looks good so far! 👍 )

PR #225, opened 2023-03-08
SCHED-331: Removed has method, replaced with __contains__.

sraaphorst:
Instead of `Plan.has`, we use `__contains__`.

Instead of doing:
```python
plan.has(obs)
```
or
```python
not plan.has(obs)
```
we can now do:
```python
obs in plan
```
and
```python
obs not in plan
```


stroncod (approved):
Great!

PR #226, opened 2023-03-10
Sched 325: new query to schedule

stroncod:


scheduler/core/service/service.py
```diff
@@ -39,6 +39,7 @@ def __call__(self):
 
         # Save to database
         PlanManager.set_plans(plans)
+        return plans
```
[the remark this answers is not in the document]
sraaphorst: (I was reading about that... falling behind on these PRs.)

scheduler/db/dbmanager.py
```diff
@@ -23,21 +24,42 @@ class DBManager:
...
+             site: Optional[FrozenSet[Site]]=None) -> List[SPlans]:
         with locking(f'{self.db_path}.lock', LOCK_SH):
             with shelve.open(self.db_path) as db:
-                if 'plans' not in db:
-                    db['plans'] = []
-                return db['plans']
-
-    def write(self, plans: List[SPlans]) -> NoReturn:
+                if start_date and end_date and site:
+                    plans_by_site = db[start_date+end_date]
```
sraaphorst, line 34: Not quite sure why you add the `start_date` and the `end_date`? I know if one (or both) are `None`, it's going to be problematic, but are you using the concatenation as an identified for plans?

scheduler/db/dbmanager.py
```diff
@@ -23,21 +24,42 @@ class DBManager:
...
+                    if Site.GS in site and Site.GN in site:
+                        return plans_by_site['both']
+                    elif Site.GN in site:
+                        return plans_by_site['GN']
+                    elif Site.GS in site:
+                        return plans_by_site['GS']
+                else:
+                    return db['plans']
```
sraaphorst, line 42: So if one of `start_date`, `end_date`, or `Site` are `None`, you return `db['plans']`?

I think maybe I'm kind of confused as to what we're doing here.

scheduler/db/dbmanager.py
```diff
@@ -23,21 +24,42 @@ class DBManager:
...
-                db['plans'] = plans
+                if start_date and end_date and site:
+                    if Site.GS in site and Site.GN in site:
+                        db[start_date+end_date]= {'both': plans}
+                    elif Site.GN in site:
+                        db[start_date+end_date]= {'GN': plans}
+                    elif Site.GS in site:
+                        db[start_date+end_date]= {'GS': plans}
+                else:
+                    db['plans'] = plans
```
sraaphorst, line 58: Okay, this clarifies it. I realize this is just a temporary measure, so it seems like it should be fine for local testing. We'll have the Heroku storage meeting tomorrow.

scheduler/db/dbmanager.py
```diff
@@ -23,21 +24,42 @@ class DBManager:
...
 
     '''
-    
-   
+
+
     def cas(self, old_db: Dict, new_db):
```
sraaphorst, line 63: What does `cas` mean?

sraaphorst (approved):
Sorry for all the rambling... just me thinking out loud.
This all looks fine to me. 👍 
We can talk more tomorrow about what we're going to do.

sraaphorst:
Maybe we should also disable dependabot (if we haven't already) and just upgrade our versions manually? That would be for another task, though.

PR #230, opened 2023-03-17
SCHED 332: Dynamic semesters based on dates.

stroncod:


scheduler/core/service/service.py
```diff
@@ -15,18 +15,28 @@
 
 
 class Service:
-    def __init__(self, start_time: Time, end_time: Time, sites: FrozenSet[Site]):
+    def __init__(self,
+                 start_time: Time,
+                 end_time: Time,
+                 semesters: FrozenSet[Semester],
```
sraaphorst, line 21: Why are the semesters now part of the `Service` class?
I'm not clear why we're moving this data around.

sraaphorst:
Ahhh... I see. It's because it doesn't make sense to put this in the config file.

PR #232, opened 2023-03-22
SCHED-333: Data Improvements.

sraaphorst:
First attempt at some improvements in model for data accessibility.

### DIFF
--- a/scheduler/core/components/optimizer/dummy.py
+++ b/scheduler/core/components/optimizer/dummy.py
@@ -44,7 +44,7 @@ class DummyOptimizer(BaseOptimizer):
         while not plans.all_done() and len(self.groups) > 0:
 
             ran_group = random.choice(self.groups)
-            if self.add(ran_group, plans, None):
+            if self.add(ran_group, plans):
                 # TODO: All observations in the group are being inserted so the whole group
                 # can be removed
                 self.groups.remove(ran_group)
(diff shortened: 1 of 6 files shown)


scheduler/core/components/optimizer/dummy.py
```diff
@@ -42,7 +44,7 @@ def _run(self, plans: Plans):
         while not plans.all_done() and len(self.groups) > 0:
 
             ran_group = random.choice(self.groups)
-            if self.add(ran_group, plans):
+            if self.add(ran_group, plans, None):
```
stroncod, line 47: Better do:
```def add(self, group: GroupData, plans: Plans, interval: Optional[Interval] = None):```
so we don't need to pass None as a parameter.

File: scheduler/core/plans/__init__.py [hunk 2]
```diff
@@ -69,6 +68,10 @@ def __init__(self, night_events: Mapping[Site, NightEvents], night_idx: int):
                                         site,
                                         len(ne.times[night_idx]))
 
+    @property
+    def time_slot_length(self):
+        return 0
```
stroncod, line 73: Why return 0 here?

stroncod (approved):

scheduler/core/components/optimizer/dummy.py
sraaphorst: I was going back and forth trying to decide. Either way works for me, so I'll give it the default parameter.

scheduler/core/plans/__init__.py
sraaphorst: Ooops. I have no idea why I did this. *doh* I'll fix this.

sraaphorst:
@stroncod I redid this PR because for some reason, it was breaking a test case and I couldn't rebase and get it to work. I went ahead and merged it and I'm closing this one.

PR #236, opened 2023-03-30
SCHED-303: Add Night Stats and Plans Summary

stroncod:
Add NightStats: a compilation of different information that it would be display in the UI.
And Plans summary: a summary by program of all nights that would also be display in the UI

This adds modifications to the Optimizer that later would need to be reconcile after GreedyMax work is completed. Probably a Dummy refactory would be in place to use common logic.

### DIFF
--- a/scheduler/core/plans/__init__.py
+++ b/scheduler/core/plans/__init__.py
@@ -25,7 +25,7 @@ class Visit:
 class NightStats:
     timeloss: str
     plan_score: float
-    plan_constraints: Conditions
+    plan_conditions: Conditions
     n_toos: int
     completion_fraction: Mapping[Band,int]
 
@@ -42,7 +42,6 @@ class Plan:
     time_slot_length: timedelta
     site: Site
     _time_slots_left: int
-    night_stats: NightStats = field(init=False)
 
     def __post_init__(self):
         self.visits = []
@@ -92,6 +91,7 @@ class Plans:
                                         ne.time_slot_length.to_datetime(),
                                         site,
                                         len(ne.times[night_idx]))
+        night_stats: NightStats = None
 
     def __getitem__(self, site: Site) -> Plan:
         return self.plans[site]
--- a/scheduler/core/plans/__init__.py
+++ b/scheduler/core/plans/__init__.py
@@ -1,7 +1,7 @@
 # Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
 # For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
 
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from datetime import datetime, timedelta
 from math import ceil
 from typing import NoReturn, Mapping, List, Optional
--- a/scheduler/core/plans/__init__.py
+++ b/scheduler/core/plans/__init__.py
@@ -1,12 +1,12 @@
 # Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
 # For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import datetime, timedelta
 from math import ceil
-from typing import NoReturn, Mapping
+from typing import NoReturn, Mapping, List, Optional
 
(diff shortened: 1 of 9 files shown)


scheduler/core/plans/__init__.py
```diff
@@ -1,12 +1,12 @@
 # Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
 # For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
```
sraaphorst, line 4: It doesn't look like you're using `field` here from what I can tell? I don't see it anywhere else in the file.

scheduler/core/service/service.py
```diff
@@ -68,3 +76,90 @@ def build_scheduler(start: Time = Time("2018-10-01 08:00:00", format='iso', scal
...
+        if program.band is Band.Band1:
+            band_1+=1
+        elif program.band is Band.Band2:
+            band_2+=1
+        elif program.band is Band.Band3:
+            band_3+=1
+        elif program.band is Band.Band4:
+            band_4+=1
+
+    return band_1, band_2, band_3, band_4
```
sraaphorst, line 99: This seems a little rough, especially returning `Tuple[int, int, int, int]`.

Could we instead do something like:
```python
# List or Set?
def _get_total_bands_by_program(programs: List[Program]) -> Dict[Band, int]:
    bands: Dict[Band, int] = {}
    for p in programs:
        bands[p.band] = bands.setdefault(p.bands, 0) + 1
    return bands
```

scheduler/graphql_mid/types.py
```diff
@@ -1,14 +1,41 @@
...
+from strawberry.scalars import JSON
 from lucupy.minimodel import ObservationID, Site, ALL_SITES
 
-from scheduler.core.plans import Plan, Plans, Visit
+from scheduler.core.plans import Plan, Plans, Visit, NightStats
+
+
+@strawberry.type
+class SNightStats:
+    """comment
```
sraaphorst, line 18: 😆

scheduler/core/service/service.py
sraaphorst: Maybe `Iterable` instead of `List` or `Set`... that seems to make more sense.

scheduler/core/service/service.py
```diff
@@ -68,3 +76,67 @@ def build_scheduler(start: Time = Time("2018-10-01 08:00:00", format='iso', scal
...
+            plan_score = 0
+
+            for visit in plan.visits:
+                obs = collector.get_observation(visit.obs_id)
+                # check if obs is a too
+                if obs.too_type is not None:
+                    n_toos += 1
+                # add to plan_score
+                plan_score += visit.score
+                # add used contrains
```
sraaphorst, line 107: Contraints 😄

scheduler/core/service/service.py
```diff
@@ -68,3 +76,67 @@ def build_scheduler(start: Time = Time("2018-10-01 08:00:00", format='iso', scal
...
+                plan_conditions.append(obs.constraints.conditions)
+                # check completition
+                program = collector.get_program(obs.belongs_to)
+
+                if program.id in all_programs_visits:
+                    all_programs_visits[program.id] += 1
+                    all_programs_scores[program.id] += visit.score
+                else:
+                    # TODO: This asssumes the observations are not spittable
```
sraaphorst, line 116: splittable

scheduler/core/service/service.py
```diff
@@ -68,3 +76,67 @@ def build_scheduler(start: Time = Time("2018-10-01 08:00:00", format='iso', scal
...
+                else:
+                    raise KeyError('Missing Band in Program!')
+        plans.night_stats =  NightStats(f'{timeloss} min',
+                                        plan_score,
+                                        Conditions.most_restrictive_conditions(plan_conditions),
+                                        n_toos,
+                                        completion_fraction)
+
+    plans_summary = {}
+    for p_id in all_programs_visits.keys():
```
sraaphorst, line 133: Not super important, but remember you iterate over the keys automatically if you do:
```python
for p_id in all_program_visits:
```
The `.keys()` is irrelevant.

sraaphorst (approved):

PR #238, opened 2023-04-04
SCHED-341: Fix filtering obs by site in collector 

stroncod:
And add sites as parameter to work with schedule query


sraaphorst (approved):
👍 
Thanks for handling this. This looks great. I'm glad you caught the `DefaultRanker` as well. I guess we should probably be more careful in our use of default parameters.

PR #246, opened 2023-04-11
SCHED-351: Ranker modifications

sraaphorst:
The `Ranker` should not be scoring all `Observation`s across all `Program`s upon creation.

Since we may create several `Ranker`s simply out of convenience (we should decide if the `Ranker` is owned by anything, e.g. possibly the `Selector`), and in subsequent scoring of `Program`s after `Observation`s or `Atom`s are executed, we do not want to use the `Collector`'s cache of `Program`s, but the modified `Program`s after GreedyMax produces a `Plan` and part of it is executed.

Formerly as well, `Observation`s that belong to observation groups were being scored, but the scoring was useless since we were lacking some information necessary to score the observation group containing the `Observation`, such as the `Conditions` / `Constraints` (see GN-2018B-Q-101-1460 and GN-2018B-Q-101-1462 in the OT for examples).

Now, the former `_score_obs` method that scored an `Observation` in the constructor for `Ranker` has been renamed to `score_observation` and is only called in the `Selector` when scoring an observation group, so observation groups missing components will not waste resources scoring their observations when they cannot even be themselves scored.

The initial `Ranker` scoring has been eliminated. All we do now is create empty arrays to be used for scoring as a convenience and so we do not have to store a reference to the `Collector` in the `Ranker`, which is unnecessary.

### DIFF
--- a/scheduler/core/components/selector/__init__.py
+++ b/scheduler/core/components/selector/__init__.py
@@ -176,7 +176,8 @@ class Selector(SchedulerComponent):
         # Calculate the group info and put it in the structure if there is actually group
         # info data inside it, i.e. feasible time slots for it in the plan.
         # This will filter out all GroupInfo objects that do not have schedulable slots.
-        unfiltered_group_data_map = self._calculate_group(program.root_group,
+        unfiltered_group_data_map = self._calculate_group(program,
+                                                          program.root_group,
                                                           sites,
                                                           night_indices,
                                                           night_configurations,
@@ -215,6 +216,7 @@ class Selector(SchedulerComponent):
         return self._group_info_map[unique_group_id]
 
     def _calculate_group(self,
+                         program: Program,
                          group: Group,
                          sites: FrozenSet[Site],
                          night_indices: NightIndices,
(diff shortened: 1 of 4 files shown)


scheduler/core/components/selector/__init__.py
```diff
@@ -176,7 +176,8 @@ def score_program(self,
         # Calculate the group info and put it in the structure if there is actually group
         # info data inside it, i.e. feasible time slots for it in the plan.
         # This will filter out all GroupInfo objects that do not have schedulable slots.
-        unfiltered_group_data_map = self._calculate_group(program.root_group,
+        unfiltered_group_data_map = self._calculate_group(program,
+                                                          program.root_group,
```
stroncod, line 180: Passing program.root_group seems redundant now if we are passing program as well.

stroncod (approved):

scheduler/core/components/selector/__init__.py
sraaphorst: It is a bit, but the method runs recursively, so we have to pass it some `Group`.
The `Program` is just passed in because we need it to score `Observation`s with info like whether or not the `Program` is a thesis program.

PR #248, opened 2023-04-12
SCHED-357: visits in output plans in chronological order

bryanmiller:
The timelines (time slots) are used to write the visits in the output plans in chronological order.


sraaphorst (approved):
👍

PR #250, opened 2023-04-20
SCHED-361: Implementation of IDs in Scheduler and cleanup

sraaphorst:
Cleanup of code and change to use of the ______ID classes.

WIP: This is not to be merged yet.

### DIFF
--- a/scheduler/core/components/optimizer/dummy.py
+++ b/scheduler/core/components/optimizer/dummy.py
@@ -37,7 +37,7 @@ class DummyOptimizer(BaseOptimizer):
                 # can be removed
                 self.groups.remove(ran_group)
             else:
-                logger.warning(f'Group {ran_group.group.unique_id()} not added')
+                logger.warning(f'Group {ran_group.group.unique_id} not added')
 
     def setup(self, selection: Selection) -> DummyOptimizer:
         """
@@ -58,7 +58,7 @@ class DummyOptimizer(BaseOptimizer):
         for observation in group.group.observations():
             plan = plans[observation.site]
             if not plan.is_full and plan.site == observation.site:
-                obs_len = plan.time2slots(observation.exec_time())
+                obs_len = plan.time2slots(plan.time_slot_length, observation.exec_time())
                 if plan.time_left() >= obs_len and observation not in plan:
                     start, start_time_slot = DummyOptimizer._first_free_time(plan)
                     visit_score = sum(group.group_info.scores[plans.night][start_time_slot:start_time_slot+obs_len])
--- a/scheduler/core/components/optimizer/greedymax.py
+++ b/scheduler/core/components/optimizer/greedymax.py
@@ -13,7 +13,7 @@ from scheduler.core.components.optimizer.timeline import Timelines
 from .base import BaseOptimizer
 from . import Interval
 
-from lucupy.minimodel import Group
+from lucupy.minimodel import Group, ObservationID, UniqueGroupID
 import numpy as np
 import numpy.typing as npt
 import matplotlib.pyplot as plt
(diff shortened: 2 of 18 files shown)


scheduler/core/components/optimizer/dummy.py
```diff
@@ -58,7 +58,7 @@ def add(self, group: GroupData, plans: Plans, interval: Optional[Interval] = Non
         for observation in group.group.observations():
             plan = plans[observation.site]
             if not plan.is_full and plan.site == observation.site:
-                obs_len = plan.time2slots(observation.exec_time())
+                obs_len = plan.time2slots(plan.time_slot_length, observation.exec_time())
```
stroncod, line 61: Why not using self then?

scheduler/core/components/optimizer/greedymax.py
```diff
@@ -13,28 +14,38 @@
...
 
-from lucupy.minimodel import Group
+from lucupy.minimodel import Group, Observation, ObservationID, Site, UniqueGroupID
 import numpy as np
 import numpy.typing as npt
 import matplotlib.pyplot as plt
 
 
+@dataclass(frozen=True)
+class ObsPlanData:
```
stroncod, line 24: This look like a `Visit`. Sounds like we are duplicating data.

stroncod (approved):
Looks good! Just a small comment.

scheduler/core/components/optimizer/dummy.py
sraaphorst: In GreedyMax, Bryan is doing this:
```
n_slots_remaining = Plan.time2slots(self.time_slot_length, time_remaining)
```
and doesn't have a `Plan` instance at that point. He was passing in `GreedyMaxOptimizer` since it has a `time_slot_length` member, so via duck typing, everything was working, but it just seemed wrong to me to fudge a `self` (which should always be a `Plan`) with a `GreedyMaxOptimizer` (which happened to have a `time_slot_length` member, so it worked).

I figured it would cause us a huge headache with mypy and typing in general.

sraaphorst: Not sure how we should handle this... any ideas? Maybe it should be an independent function that just takes a `time_slot_length` and a `time_delta`. That way, we can eliminate it from `Plan` and still use it where needed, and use it from `GreedyMax` in the way that Bryan is using it.

scheduler/core/components/optimizer/greedymax.py
```diff
@@ -403,12 +418,14 @@ def add(self, group_data: GroupData, night: int, interval: Optional[Interval] =
...
+                visit_score = sum(group_data.group_info.scores[night][start_time_slot:start_time_slot + obs_len])
 
-                self.obs_in_plan[observation.id] = {'obs': observation, 'obs_len': obs_len,
-                                                    'start': start, 'visit_score': visit_score}
+                self.obs_in_plan[observation.id] = ObsPlanData(
+                    obs=observation,
+                    obs_start=start,
+                    obs_len=obs_len,
+                    visit_score=visit_score
+                )
```
sraaphorst, line 426: With the `ObsPlanData`, I was just trying to get rid of the dictionary on the left here. ☹️ 
Since this is using `Timeline`, I don't think we have enough information yet for a `Visit` yet? I'm going to have to go over the code in finer detail, but I'm anxious to change it too dramatically until Bryan has it running correctly. I think we could manage to combine some methods and come up with something that represents `Timeline` and `Plan` so that this isn't duplicated, but right now, that may not be the best idea. 😦

scheduler/core/components/optimizer/greedymax.py
```diff
@@ -424,9 +439,17 @@ def output_plans(self, plans: Plans) -> None:
...
+                    # TODO: UniqueGroupID of an observation group. Maybe we can make obs_in_plan a map from
+                    # TODO: UniqueGroupID to... something, instead of an ObservationID to an ObsPlanData.
+                    unique_group_id = self.obs_group_ids[idx]
+                    obs_id = ObservationID(unique_group_id.id)
 
                     # Add visit to final plan
-                    plans[timeline.site].add(self.obs_in_plan[obs_id]['obs'], self.obs_in_plan[obs_id]['start'],
-                                             start_time_slot, self.obs_in_plan[obs_id]['obs_len'],
-                                             self.obs_in_plan[obs_id]['visit_score'])
+                    obs_in_plan = self.obs_in_plan[obs_id]
```
sraaphorst, line 450: @stroncod Like this is driving me nuts. We can't go from a `UniqueGroupID` to an `ObservationID` although we can go in the other direction since we know that an `Observation` will always be contained in an observation `Group` that has the same ID in its `UniqueGroupID` as is on the `Observation`'s `ObservationID`... however, the converse is not true, and we will have `UniqueGroupID`s that don't correspond to `ObservationID`... I'm wondering if we should just change the `self.obs_in_plan` mapping to use `UniqueGroupID` instead, but then it seems it should have value `Visit` instead of `ObsPlanData` as you say... although the way things are structured, we don't have enough information for that now.

PR #264, opened 2023-05-09
SCHED-374: Fixed copying issues.

sraaphorst:
This fixes an issue in the `Selector` where when a `select` was run, too many things were copied.

This ended up with things in the `Selection` no longer being linked, e.g. a `Group` in the `Selection` would not be linked to the `Program` from which it originated, and thus making any changes to the `Group` would not percolate up to the `Program` level. Similarly, this was a problem with `Observation`s and the `Group` to which they belonged.

Now,, they should be working correctly and the members of a `Selection` should now have their appropriate relationships maintained.

### DIFF
--- a/scheduler/core/components/optimizer/greedymax.py
+++ b/scheduler/core/components/optimizer/greedymax.py
@@ -4,7 +4,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


stroncod (approved):
I'll go ahead and merged it so Bryan can move forward with this.

PR #265, opened 2023-05-09
SCHED-353: Program rescoring

bryanmiller:
Rescore programs after a group is included in a plan. Only incomplete groups have their scores updates, leaving the scores of the groups in the plans as they were when selected.

### DIFF
--- a/scheduler/core/components/optimizer/greedymax.py
+++ b/scheduler/core/components/optimizer/greedymax.py
@@ -37,7 +37,7 @@ class GreedyMaxOptimizer(BaseOptimizer):
     """
 
     def __init__(self, min_visit_len: timedelta = timedelta(minutes=30), show_plots: bool = False):
-        # self.selection = Selection
+        self.selection = Selection
         self.group_data_list: List[GroupData] = []
         self.group_ids: List[UniqueGroupID] = []
         self.obs_group_ids: List[UniqueGroupID] = []
@@ -52,7 +52,7 @@ class GreedyMaxOptimizer(BaseOptimizer):
         """
         Preparation for the optimizer.
         """
-        # self.selection = selection
+        self.selection = selection
         self.group_ids = list(selection.schedulable_groups)
         self.group_data_list = list(selection.schedulable_groups.values())
         # self._process_group_data(self.group_data_list)
@@ -441,7 +441,10 @@ class GreedyMaxOptimizer(BaseOptimizer):
                 # add to timeline (time_slots)
                 start_time_slot, start = timeline.add(iobs, obs_len, best_interval)
                 # pseudo (internal) time charging
+                program = self.selection.program_info[group_data.group.program_id].program
+                print(f"{program.id.id}: total_used: {program.total_used()} program_used: {program.program_used()}")
                 self._charge_time(observation)
+                print(f"total_used after: {program.total_used()} {program.program_used()}")
 
                 # Sergio's Note:
                 # Both of these lines are added to calculate NightStats. This could be modified,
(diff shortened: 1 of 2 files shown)


scheduler/core/components/optimizer/greedymax.py
```diff
@@ -37,7 +37,7 @@ class GreedyMaxOptimizer(BaseOptimizer):
     """
 
     def __init__(self, min_visit_len: timedelta = timedelta(minutes=30), show_plots: bool = False):
-        # self.selection = Selection
+        self.selection = Selection
```
sraaphorst, line 40: This isn't quite right, since you're setting `self.selection` to a type (`Selection`). It would be like doing:

```
self.index = int
```

Instead, what we should do is:
```
self.selection: Optional[Selection] = None
```
(saying that `self.selection` will be a `Selection`, but right now, we don't have the value for it).

Then when you set it in `setup`, it looks good.

scheduler/core/components/optimizer/greedymax.py
```diff
@@ -374,6 +374,26 @@ def _charge_time(observation: Observation, atom_start: int = 0, atom_end: int =
...
+            group_data = program_calculations.group_data_map[unique_group_id]
+            group, group_info = group_data
+            schedulable_group = self.selection.schedulable_groups[unique_group_id]
+            print(f"{unique_group_id} {schedulable_group.group.exec_time()} {schedulable_group.group.total_used()}")
+            print(f"\tOld max score: {np.max(schedulable_group.group_info.scores[0]):7.2f} new max score[0]: "
+                  f"{np.max(group_info.scores[0]):7.2f}")
+            # update scores in schedulable_groups if the group is not completely observed
+            if schedulable_group.group.exec_time() >= schedulable_group.group.total_used():
+                for ii in range(len(group_info.scores)):
+                    schedulable_group.group_info.scores[ii] = group_info.scores[ii]
```
sraaphorst, line 394: It seems like you shouldn't have to do all this work here... we can leave it for now, but I'm going to look at making the rescoring handle this detail.

In any case, it would be much easier to just set the whole `schedulable_groups.group_info.scores` in one fell swoop rather than loop over the elements.

Instead of:

```python
if schedulable_group.group.exec_time() >= schedulable_group.group.total_used():
    for ii in range(len(group_info.scores)):
        schedulable_group.group_info.scores[ii] = group_info.scores[ii]
```

I think this would work:

```python
if schedulable_group.group.exec_time() >= schedulable_group.group.total_used():
    schedulable_group.group_info.scores = group_info.scores
```

And if not that, then this:

```python
if schedulable_group.group.exec_time() >= schedulable_group.group.total_used():
    schedulable_group.group_info.scores = group_info.scores[:]
```

scheduler/core/components/optimizer/greedymax.py
```diff
@@ -374,6 +374,26 @@ def _charge_time(observation: Observation, atom_start: int = 0, atom_end: int =
             observation.sequence[n_atom].observed = True
             observation.sequence[n_atom].qa_state = QAState.PASS
 
+    def _update_score(self, program) -> None:
+        """Update the scores of the incomplete groups in the scheduled program"""
+
+        print("Call score_program")
+        program_calculations = self.selection.score_program(program)
```
sraaphorst, line 381: Because of the initial inferred type of `self.selection` as I mentioned above, this generates a warning. When we change the constructor `self.simulation` line, that warning goes away.

scheduler/core/components/optimizer/greedymax.py
```diff
@@ -374,6 +374,26 @@ def _charge_time(observation: Observation, atom_start: int = 0, atom_end: int =
             observation.sequence[n_atom].observed = True
             observation.sequence[n_atom].qa_state = QAState.PASS
 
+    def _update_score(self, program) -> None:
```
sraaphorst, line 377: Can we add type information here?
```python
def _update_score(self, program: Program) -> None:
    ...
```

sraaphorst (commented):
Looks good!
One small typing error for `self.selection`, and a few little details.

scheduler/core/components/optimizer/greedymax.py
bryanmiller: When I try one of your options pycharm says "'GroupInfo' object attribute 'scores' is read-only". I tried running it anyway but I get the error "dataclasses.FrozenInstanceError: cannot assign to field 'scores'"

bryanmiller: but this does work
            if schedulable_group.group.exec_time() >= schedulable_group.group.total_used():
                schedulable_group.group_info.scores[:] = group_info.scores[:]

sraaphorst: Let's leave it as is for now, and I'll unfreeze the scores... then we can get rid of the loop later.
There's really no rush to do so.

sraaphorst (approved):
Great!

scheduler/core/components/optimizer/greedymax.py
sraaphorst: Yeah, that will work fine. I'll delay holding off changing the frozen status of scores until we have everything working.

PR #266, opened 2023-05-11
SCHED 373: Add Instrument to schedule query

stroncod:


scheduler/core/components/optimizer/dummy.py
```diff
@@ -62,7 +62,7 @@ def add(self, group: GroupData, plans: Plans, interval: Optional[Interval] = Non
                 if plan.time_left() >= obs_len and observation not in plan:
                     start, start_time_slot = DummyOptimizer._first_free_time(plan)
                     visit_score = sum(group.group_info.scores[plans.night][start_time_slot:start_time_slot+obs_len])
-                    plan.add(observation, start, start_time_slot, obs_len, visit_score)
+                    plan.add(observation, start, start_time_slot, obs_len, visit_score, observation)
```
sraaphorst, line 65: 😄 Why do we pass `observation` twice to `plan.add`? It seems like we should be passing an `Optional[Resource]` here based on the method signature?

sraaphorst: ...or maybe I am just confused... 😓

stroncod: My bad. I'll fix it.

sraaphorst (approved):
👍

PR #267, opened 2023-05-17
SCHED-378: Fixing OcsResourceService issues.

sraaphorst:
There were some issues in `OcsResourceServices`:
1. Using `date` (a type) as a variable name.
2. Conversion of single items into `FrozenSet`.
3. Type information for `EngTask` and `Fault` should be `Set`.

In the last case, this would be better as `FrozenSet`, but at this point, the conversion would be difficult.


stroncod (approved):
Thanks for the fix ups!

PR #277, opened 2023-07-17
SCHED-404: First cleanup

stroncod:


scheduler/core/components/optimizer/base.py
```diff
@@ -2,9 +2,9 @@
 # For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
 
 from abc import ABC, abstractmethod
-from datetime import datetime
-from typing import Mapping, List, Optional, Tuple
-
+from dataclasses import dataclass
+from datetime import timedelta
+from typing import Mapping, List, Optional, Union
```
sraaphorst, line 7: Just a reminder: we can omit `Union` from now on and just use `|` as in `int | str` for example.

sraaphorst (approved):
Looks like a good start to the cleanup! 👍

PR #280, opened 2023-07-26
SCHED 407: Time accounting

stroncod:


scheduler/core/components/collector/__init__.py
```diff
@@ -481,3 +482,31 @@ def night_configurations(self,
...
+                    atom.program_used = atom.prog_time
+                    atom.partner_used = atom.part_time
+
+                    # Charge acquisition to the first atom
+                    if atom_idx == 0:
+                        if observation.obs_class == ObservationClass.PARTNERCAL:
+                            atom.program_used += observation.acq_overhead
+                        elif (observation.obs_class == ObservationClass.SCIENCE or
+                              observation.obs_class == ObservationClass.PROGCAL):
+                            atom.program_used += observation.acq_overhead
```
sraaphorst, line 509: Instead of doing this, you should just be able to move this out of the for loop and do it directly for `first_atom = observation.sequence[0]`.

scheduler/core/components/collector/__init__.py
```diff
@@ -481,3 +482,31 @@ def night_configurations(self,
...
+
+                    atom.observed = True
+                    atom.qa_state = QAState.PASS
```
sraaphorst, line 512: 👍

sraaphorst (commented):
A few small changes and a question.

I'm not fully sure this is the right logic, but apart from the ONGOING, it looks like what Bryan is doing, so let's go with this.

scheduler/core/components/collector/__init__.py
stroncod: The problem is the for has to be changed as well because it would replace the values in the first iteration and if the first atom is skip entirely then now we have to add the first atom.progtime as well. I guess is just matter of personal preference?

sraaphorst (approved):
Let's merge it!

scheduler/core/components/collector/__init__.py
bryanmiller: if atom_idx == 0 should be if atom_idx = v.atom_start_idx.

PR #296, opened 2023-08-25
Add note parsing for 'do not split' info, bugfixes

bryanmiller:
Parse notes while traversing the observation tree to look for text indicting that observations should not be split. The note is assumed to apply to all observations at or below its level (program, group, observation). The atom generation then uses this information. Tested with GS-2018B-Q-108-18.
I also fixed bugs related to note parsing for FT scheduling info (all top-level notes need to be searched) and to atom generation.


sraaphorst (approved):
This looks very good. 👍  Let's go ahead and get it merged.

I might add a test case for it because we've been trying to keep moving forward and the lack of test cases has been an issue. I'll message you about that

PR #303, opened 2023-08-29
Fix setting observing mode from atom configs

bryanmiller:
Setting and use of obs_mode by GreedyMax
atom bugfixes
add MOS mask barcodes to resources, so the selector should now select a MOS observation only if the MOS mask is in the instrument.

Requires the following PR
https://github.com/gemini-hlsw/lucupy/pull/60

### DIFF
--- a/scheduler/services/resource/base.py
+++ b/scheduler/services/resource/base.py
@@ -8,7 +8,7 @@ from datetime import date, datetime, timedelta
 from typing import Dict, List, Set, Tuple, Union, Final
 from io import BytesIO, StringIO
 
-from lucupy.minimodel import Site, ALL_SITES
+from lucupy.minimodel import Site, ALL_SITES, Resource
 from lucupy.helpers import str_to_bool
 import gelidum
 import requests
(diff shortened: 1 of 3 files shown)


scheduler/services/resource/base.py
```diff
@@ -102,6 +102,17 @@ def __init__(self, sites: FrozenSet[Site] = ALL_SITES):
         # The final output from this class: the configuration per night.
         self._night_configurations: Dict[Site, Dict[date, NightConfiguration]] = {site: {} for site in self._sites}
 
+    @staticmethod
+    def _mdf_to_barcode(mdfname: str, inst: str) -> Resource:
+        """Legacy MOS mask barcode convention"""
+        barcode = None
+        instd = {'GMOS': '1', 'GMOS-N': '1', 'GMOS-S': '1', 'Flamingos2': '3'}
+        semd = {'A': '0', 'B': '1'}
+        progd = {'Q': '0', 'C': '1', 'L': '2', 'F': '3', 'S': '8', 'D': '9'}
```
sraaphorst, line 111: Since these are constants, I'm going to move them out of the method and into the class, but I can take care of that after.

scheduler/services/resource/base.py
```diff
@@ -102,6 +102,17 @@ def __init__(self, sites: FrozenSet[Site] = ALL_SITES):
...
+        if inst in instd.keys():
+            barcode = instd[inst] + semd[mdfname[6]] + progd[mdfname[7]] + mdfname[-6:-3] + mdfname[-2:]
```
sraaphorst, line 113: Also, I'll make this into an f-string instead of adding strings together.

sraaphorst (approved):
There's a couple very small changes I want to make, but I'll make a tiny PR after this is merged. I left them as part of the review so I would remember to do it.

Let's get it into the repo.

PR #306, opened 2023-08-31
SCHED 450: Builder by mode instead of config file

stroncod:
Following the architecture design, the different components of the scheduler are need to be build according to the mode the system is currently on. Also different sessions can use different modes (except for Operation) so this is important to handle dynamically instead of in a fixed config file. 

This uses previously Builder class to make the distinction between modes and the logic behind this. Currently one of the test fails because the load_program method is on core service object and that is going to change in further developments. Demo and run_greedymax.py works as suppose to.


scheduler/graphql_mid/schema.py
```diff
@@ -67,12 +61,22 @@ async def load_sources_files(self, files_input: UseFilesSourceInput) -> SourceFi
...
     @strawberry.mutation
-    def change_origin(new_origin: SOrigin) -> ChangeOriginSuccess:
-        old = str(builder.sources.origin)
+    def change_origin(self, new_origin: SOrigin, mode: SchedulerModes) -> ChangeOriginSuccess:
+
+        old = str(sources.origin)
+        new = str(new_origin)
+        if new == 'OCS' and mode is SchedulerModes.SIMULATION:
+            raise ValueError('Simulation mode can only work with GPP origin source.')
+        elif new == 'GPP' and mode is SchedulerModes.VALIDATION:
```
sraaphorst, line 75: It would be good to have the magic strings 'OCS' and 'GPP' defined elsewhere since they're not only used here but in asserts, but it's not the end of the world.

sraaphorst (approved):
This looks good. I had one small suggestion, but approved! 👍

PR #308, opened 2023-09-06
HOTFIX: Quantity vs numpy array

stroncod:
Strips wind sep and longitude from Quantity and uses regular numpy array


sraaphorst (approved):
Thanks for changing this! 👍

PR #309, opened 2023-09-07
SCHED-454: Fixed issue where illegal numpy broadcasts were being done.

sraaphorst:
It turns out that in the `Selector.selection` method, `ALL_SITES` was always being used, which was causing errors since if only one site was given to the `Collector` unless the `NightEvent` information for the nights desired had already been computed and cached in `NightEventsManager`, we had observations being evaluated using information that didn't exist in `NightEventsManager` against env data, which generated numpy broadcast errors in the size of arrays.

Note that this also adds a script that can be run to test to make sure things work now (`run_greedymax_twice.py`) so as to run the same scenario that was originally discovered by German in Mercury without all the additional noise Mercury generates.

There is also some error-handling code added, and the scripts `run_greedymax.py` and `run_greedymax_twice.py` now both properly calculate the `num_nights_to_schedule` programmatically from the start and the end of the scheduling period instead of using a fixed value.

Finally, in the Mercury notebook, I added an import for `astropy` `Time`, since there was a crash where it couldn't find the definition.

### DIFF
--- a/scheduler/core/components/selector/__init__.py
+++ b/scheduler/core/components/selector/__init__.py
@@ -57,7 +57,7 @@ class Selector(SchedulerComponent):
                              'Cannot proceed.')
 
     def select(self,
-               sites: FrozenSet[Site] = ALL_SITES,
+               sites: Optional[FrozenSet[Site]] = None,
                night_indices: Optional[NightIndices] = None,
                ranker: Optional[Ranker] = None) -> Selection:
         """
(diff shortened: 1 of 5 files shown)


scheduler/core/components/selector/__init__.py
```diff
@@ -79,6 +79,11 @@ def select(self,
 
         An AND group must be able to perform all of its children.
         """
+        if sites is None:
```
stroncod, line 82: maybe we could fetch sites always instead of select them as parameters

scheduler/core/components/selector/__init__.py
```diff
@@ -265,8 +271,10 @@ def _calculate_observation_group(self,
         if obs.status not in {ObservationStatus.ONGOING, ObservationStatus.READY}:
             logger.warning(f'Selector skipping observation {obs.id}: status is {obs.status.name}.')
             return group_data_map
+
+        # This should never happen.
         if obs.site not in sites:
-            logger.warning(f'Selector skipping observation {obs.id}: not at a designated site.')
+            logger.error(f'Selector requested to score {obs.id}: not at a designated site.')
             return group_data_map
```
stroncod, line 278: This should raise something right?

stroncod (approved):
Great fix! Approved.

PR #318, opened 2023-09-21
Slight modifications: copyright and couple changes to stats calcs

sraaphorst:
* Add copyright header
* Use `Counter[Band]` instead of `Dict[Band, int]`
* `sum` over generator comprehension instead of `for` loop
* PyCharm did some spacing formatting
* Isolated running code to main for chronicle, which we will be changing anyway

### DIFF
--- a/scheduler/core/statscalculator/__init__.py
+++ b/scheduler/core/statscalculator/__init__.py
@@ -2,7 +2,7 @@
 # For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
 
 from collections import Counter
-from typing import Dict, List
+from typing import Dict, List, Tuple
 
 from lucupy.minimodel import Band, Conditions, ProgramID
 from lucupy.types import ZeroTime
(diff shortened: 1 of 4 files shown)


scheduler/core/statscalculator/__init__.py
```diff
@@ -58,21 +59,17 @@ def calculate_plans_stats(all_plans: List[Plans],
...
-                completion_fraction = {b: 0 for b in Band}
 
         plans_summary = {}
         for p_id in all_programs_scores:
             program = collector.get_program(p_id)
             total_used = program.total_used()
-            prog_total = timedelta()
-            for o in program.observations():
-                prog_total += (o.part_time() + o.acq_overhead + o.prog_time())
+            prog_total = sum((o.part_time() + o.acq_overhead + o.prog_time() for o in program.observations()),
```
stroncod, line 67: Much more clean!

stroncod (approved):

PR #322, opened 2023-09-28
SCHED-392: Event Handling

stroncod:
For discussion on event handling for the scheduler.


scheduler/core/eventsqueue/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .queue import EventQueue
+from .events import (Event, Interruption, Blockage,
+                     WeatherChange, ResumeNight, Fault,
+                     Rto0s)
```
sraaphorst, line 4: Why the `Rto0s`, out of curiosity, instead of `Rtoos` or something?

Let's not worry about rapid ToOs just yet, and focus on one event... I think RToOs should come in as events that are either new programs or changes to programs, but we should discuss this and just get the event model working first.

scheduler/core/meta/__init__.py
```diff
@@ -11,3 +14,13 @@ def __call__(cls, *args, **kwargs):
...
+
+
+@dataclass
+class AbstractDataclass(ABC):
+    pass
+
+
+@dataclass(frozen=True)
+class FrozenAbstractDataclass(ABC):
+    pass
```
sraaphorst, line 26: Not sure why we need these? Why wouldn't we just pass frozen or non-frozen dataclasses in? I could be missing something.

scheduler/core/eventsqueue/events.py
```diff
@@ -0,0 +1,68 @@
...
+
+from scheduler.core.meta import AbstractDataclass, FrozenAbstractDataclass
+
+
+@dataclass(frozen=True)
+class Interruption(FrozenAbstractDataclass):
+    """
+    Parent class for any interruption in the night that would
+    cause a new schedule to be created.
+    """
```
sraaphorst, line 15: This probably should be:
```
Parent class for any interruption that could cause the generation of a new plan.
```

The idea is for these events to be checked and determine if we should create a new plan based on the information provided.

scheduler/core/eventsqueue/events.py
```diff
@@ -0,0 +1,68 @@
...
+    start: datetime
```
sraaphorst, line 16: Thinking out loud: it seems that `start` should maybe be a time slot and not a `datetime`.

It probably doesn't matter because I think we can convert between the two, but we should make sure we have the functionality to do so.

scheduler/core/eventsqueue/events.py
```diff
@@ -0,0 +1,68 @@
...
+    reason: str
+    end: datetime = None  # needs a resume event
+
+    def ends(self, end: datetime) -> None:
+        self.end = end
+
+    def time_loss(self) -> timedelta:
+        if self.end:
```
sraaphorst, line 37: Maybe this should be to the end of the night if the Blockage is not resolved? Again, just thinking out loud... the fact that these are still questions that we haven't resolved or discussed makes me think we might need a clearer direction moving forward.

scheduler/core/eventsqueue/events.py
```diff
@@ -0,0 +1,68 @@
...
+class ResumeNight(Interruption):
+    """
+    Event that let the scheduler knows that the night can be resumed
+    by a new
+    """
+    pass
+
+
+class Fault(Blockage):
+    affects: List[Resource]
```
sraaphorst, line 55: Probably a `FrozenSet`, and shouldn't this be a `dataclass`?

scheduler/core/eventsqueue/queue.py
```diff
@@ -0,0 +1,38 @@
...
+        else:
+            self._add(events, night_idx)
+
+    def check_blockage(self, resume_event: ResumeNight) -> Blockage:
+        if self._blockage_stack and len(self._blockage_stack) == 1:
+            b = self._blockage_stack.pop()
+            b.ends(resume_event.start)
+            return b
+
+        raise RuntimeError('Missing blockage for ResumeNight')
```
sraaphorst, line 35: I'm not quite sure how this will work.

sraaphorst (commented):
Some general notes:

- The complication involved in `Blockage` events is why I think we should focus on one simple event instead of on all possible events for now. We can worry about blocking events once we know that weather events work.
- Do we need to worry about the Strawberry side of things just yet? I would think that we can start by focusing on internal events first.
- I think this dives into the details a little too deeply. We should probably be discussing this more so that we follow a well-formulated design instead of implementing too much all at once, and make incremental changes with small PRs that only change minimal things.
- I'll go over the script in my morning.

scheduler/core/eventsqueue/__init__.py
stroncod: Agree!

scheduler/core/meta/__init__.py
stroncod: The idea is to make `Interruption` and `Blockage` abstract. That could be done just using ABC on those classes too.

scheduler/core/eventsqueue/events.py
stroncod: I was thinking the user would chose a time in the night instead of a time slot.

scheduler/core/eventsqueue/events.py
stroncod: That's probably true, I did not add to much logic on it because we were focusing on weather changes. I think for Validation/Simulation we are going to know the end event. The problem is if the event last for more than one night. I agree we should be clearer on that.

scheduler/core/eventsqueue/events.py
sraaphorst: Ah... that makes sense if the data is coming from a user or from an OCS service (which will have `datetime` labels).

I'm wondering, though, since the Scheduler works on time slots, if we should translate it into a time slot so when we get to time slot `ts_idx`, we can look to see if there are any events, and if there are, then we can grab them, send them to the Scheduler, and check if they should trigger the generation of a new plan.

scheduler/core/eventsqueue/events.py
sraaphorst: Let's not worry too much about this now, and focus on weather changes specifically. Then once we make sure that those are working, we can decide how we want to model other types of events.

sraaphorst:
I did an experiment on dataclass inheritance just to see what would happen:

```python
@dataclass(frozen=True)
class AbstractFrozenDataclass(ABC):
    pass

@dataclass
class ImplementedFrozenDataclass(AbstractFrozenDataclass):
    a: int
    b: str

@dataclass(frozen=True)
class FrozenDataclass:
    a: int
    b: str
```

Then, the following code works fine (the `frozen` doesn't carry down through the inheritance):

```python
c1 = ImplementedFrozenDataclass(a=5, b='hello')
c1.a = 10
```

The following code raises a `dataclasses.FrozenInstanceError`:
```python
c2 = FrozenDataclass(a=5, b='hello')
c1.a = 10
# dataclasses.FrozenInstanceError: cannot assign to field 'a'
```

But then, we also get this interesting behavior:

```python
@dataclass(frozen=True)
class MyAbstractDataclass(ABC):
    a: int

@dataclass
class MyConcreteDataclass(MyAbstractDataclass):
    b: str
# TypeError: cannot inherit non-frozen dataclass from a frozen one
```

so unless we have a trivially empty dataclass, `frozen` must propagate through the subclasses, and if we have a trivially empty dataclass, marking it `frozen` has no effect on its subclasses.

I was curious, so I played around a bit. 🙃

scheduler/scripts/run_with_events.py
```diff
@@ -0,0 +1,126 @@
...
+        ['Q', 'LP', 'FT', 'DD'],
+        1.0
+    )
+    queue = EventQueue()
+    weather_change = WeatherChange(new_conditions=Conditions(iq=ImageQuality.IQ20,
+                                                             cc=CloudCover.CC50,
+                                                             sb=SkyBackground.SBANY,
+                                                             wv=WaterVapor.WVANY),
+                                   start=datetime(2018, 10, 1, 10),
+                                   reason='Worst image quality',
```
sraaphorst, line 45: This is backwards. `IQ20` is the best image quality, and `IQANY` is the worst image quality. (Anything can be scheduled in `IQ20`, but in `IQANY`, only things with `IQANY` can be scheduled.)

stroncod:
> I did an experiment on dataclass inheritance just to see what would happen:
> 
> ```python
> @dataclass(frozen=True)
> class AbstractFrozenDataclass(ABC):
>     pass
> 
> @dataclass
> class ImplementedFrozenDataclass(AbstractFrozenDataclass):
>     a: int
>     b: str
> 
> @dataclass(frozen=True)
> class FrozenDataclass:
>     a: int
>     b: str
> ```
> 
> Then, the following code works fine (the `frozen` doesn't carry down through the inheritance):
> 
> ```python
> c1 = ImplementedFrozenDataclass(a=5, b='hello')
> c1.a = 10
> ```
> 
> The following code raises a `dataclasses.FrozenInstanceError`:
> 
> ```python
> c2 = FrozenDataclass(a=5, b='hello')
> c1.a = 10
> # dataclasses.FrozenInstanceError: cannot assign to field 'a'
> ```
> 
> But then, we also get this interesting behavior:
> 
> ```python
> @dataclass(frozen=True)
> class MyAbstractDataclass(ABC):
>     a: int
> 
> @dataclass
> class MyConcreteDataclass(MyAbstractDataclass):
>     b: str
> # TypeError: cannot inherit non-frozen dataclass from a frozen one
> ```
> 
> so unless we have a trivially empty dataclass, `frozen` must propagate through the subclasses, and if we have a trivially empty dataclass, marking it `frozen` has no effect on its subclasses.
> 
> I was curious, so I played around a bit. 🙃

I got the same results and that is why there is two because dataclass cant inherit from frozen. That is why the decorator also has to be on the child as well.

sraaphorst:
Ahhh... okay, now I think I understand better what your intention was with the dataclasses... I was confused, but this makes more sense to me now. Thanks for explaining.

scheduler/scripts/run_with_events.py
stroncod: Thanks, my lack of astronomy was showing

scheduler/core/components/selector/__init__.py
```diff
@@ -244,6 +244,10 @@ def score_program(self,
             unfiltered_group_data_map=unfiltered_group_data_map
         )
 
+    def update_conditions(self, new_conditions: Conditions)-> None:
+        self.default_cc = new_conditions.cc
+        self.default_iq = new_conditions.iq
```
sraaphorst, line 249: Did we want to change this per site?

scheduler/core/meta/__init__.py
sraaphorst: Since this does nothing, we should probably get rid of these two and use the ABC route.

scheduler/graphql_mid/schema.py
```diff
@@ -79,6 +83,17 @@ def change_origin(self, new_origin: SOrigin, mode: SchedulerModes) -> ChangeOrig
...
         return ChangeOriginSuccess(from_origin=old, to_origin=str(new_origin))
 
+    @strawberry.mutation
+    def add_events(self, events_input: AddEventInput) -> EventsAddedResponse:
+
+        for e in events_input.events:
+            match e:
+                case isinstance(e, NewWeatherChange):
+                    event_queue.add_events(e.to_scheduler_event())
+                    return EventsAddedSuccess(True, 'Weather change')
```
sraaphorst, line 93: This could be good as an example. Are we going to add it to the UI?

scheduler/scripts/run_with_events.py
sraaphorst: Remember to change this. I noticed it in the output yesterday. 😄

scheduler/scripts/run_with_events.py
```diff
@@ -0,0 +1,131 @@
...
+                                   )
+    queue.add_events(weather_change, 0)
+
+    builder = ValidationBuilder(Sources(), queue)
+
+    start = Time("2018-10-01 08:00:00", format='iso', scale='utc')
+    end = Time("2018-10-03 08:00:00", format='iso', scale='utc')
+    num_nights_to_schedule = int(round(end.jd - start.jd)) + 1
```
sraaphorst, line 55: We should probably get rid of this since `num_nights_to_schedule` is no longer calculated to be the difference between `start` and `end`: see `run_greedymax.py` as an example.

scheduler/scripts/run_with_events.py
```diff
@@ -0,0 +1,131 @@
...
+        # Run eventless timeline
+        selection = selector.select(night_indices=night_indices)
+        # Run the optimizer to get the plans for the first night in the selection.
+        plans = optimizer.schedule(selection)
+        night_timeline.add(night_idx, 0, Twilight(start.to_datetime()), plans[0])
+
+        if events_by_night:
+            while events_by_night:
+                event = events_by_night.pop()
+                event_start_time_slot = ceil((event.start - start.to_datetime()).total_seconds()/60)
```
sraaphorst, line 110: This assumes a timeslot length of one minute... the timeslot length should be in the `Collector` and we should maybe make a method in there (or an independent method, preferably, that takes a `timedelta` and a `timeslot_length: timedelta`) and converts and use that.

scheduler/scripts/run_with_events.py
```diff
@@ -0,0 +1,131 @@
...
+                if isinstance(event, WeatherChange):
+                    selector.default_iq = event.new_conditions.iq
+                    selector.default_cc = event.new_conditions.cc
```
sraaphorst, line 113: Good, and okay for now, but I think we need to start thinking of moving all of this logic out of a script and into an object.

sraaphorst (commented):
I'll take one more look at this once I'm up for the day... but so far, it is looking pretty good!
Before we merge, I'd like to take care of some of the things discussed... if we don't get it merged before then, we can run a demo off your branch, I think.

PR #323, opened 2023-10-03
Fixes to internal time accounting, avoid multiple visits on one night

bryanmiller:
This fixes some bugs in the greedymax internal time accounting. Unused partner_cal (telluric) observations are set to Inactive, avoiding errors during rescoring. Any group added to a plan is removed from group_data_list so that it won't be scheduled again. This is a simple way to avoid multiple visits of an observation on a single night.

### DIFF
--- a/scheduler/core/components/optimizer/greedymax.py
+++ b/scheduler/core/components/optimizer/greedymax.py
@@ -777,7 +777,8 @@ class GreedyMaxOptimizer(BaseOptimizer):
                     # print(f"Group {max_group_info.group_data.group.unique_id.id} with "
                     #      f"max score {max_group_info.max_score} added.")
                     # Remove group from list if completely observed
-                    if max_group_info.group_data.group.program_used() >= max_group_info.group_data.group.prog_time():
+                    # if max_group_info.group_data.group.program_used() >= max_group_info.group_data.group.prog_time():
+                    if max_group_info.group_data.group.total_used() >= max_group_info.group_data.group.exec_time():
                         self.group_data_list.remove(max_group_info.group_data)
             else:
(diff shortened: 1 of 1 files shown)


bryanmiller:
Depends on lucupy changes in https://github.com/gemini-hlsw/lucupy/pull/65

scheduler/core/components/optimizer/greedymax.py
```diff
@@ -628,7 +628,7 @@ def _charge_time(observation: Observation, atom_start: int = 0, atom_end: int =
             atom_end += seq_length
 
         # Update observation status
-        if atom_end == seq_length:
+        if atom_end == seq_length - 1:
```
sraaphorst, line 631: So this explains why we were getting strange results with observations being marked `READY` despite the fact that they were complete?

scheduler/core/components/optimizer/greedymax.py
```diff
@@ -923,9 +924,16 @@ def add(self, night_idx: NightIndex, max_group_info: GroupData | MaxGroup) -> bo
...
+            # Inactivate any standards not used
+            for obs in part_obs:
+                if obs not in standards:
+                    print(f'Deactivating observation {obs.id.id}')
+                    obs.status = ObservationStatus.INACTIVE
+
             # TODO: Shift to remove any gaps in the plan?
 
             # Re-score program (pseudo time accounting)
+            print(f'Update score')
```
sraaphorst, line 936: We should probably eliminate the debugging print here, but I'll put that in my next PR.

sraaphorst (approved):
Great. Thanks for finding and fixing this.

PR #346, opened 2023-10-30
SCHED-510: Fix NumSlots 

stroncod:
+++++ NIGHT 1 +++++
Plan for site: GS
	Execution time                          ObsID                   Score  StartAtom    EndAtom  StartSlot    EndSlot   NumSlots
	2018-09-30 20:55 to 2018-09-30 22:01:   GS-2018B-Q-104-21      373.97          0          1         19          85        66
	2018-09-30 21:44 to 2018-09-30 22:36:   GS-2018B-Q-115-11     3436.31          0         19         68         120        52
	2018-09-30 22:36 to 2018-10-01 00:04:   GS-2018B-Q-115-11      395.58          0          5        120         208        88
	2018-09-30 23:52 to 2018-10-01 04:10:   GS-2018B-Q-115-7       976.35          0          5        196         454       258
	2018-10-01 04:10 to 2018-10-01 04:18:   GS-2018B-Q-115-10       24.74          0          1        454         462         8
	2018-10-01 04:17 to 2018-10-01 05:49:   GS-2018B-Q-105-81      944.98          0          5        461         553        92
	2018-10-01 05:48 to 2018-10-01 06:31:   GS-2018B-Q-223-31      834.14          0         23        552         595        43
DONE

new output

### DIFF
--- a/scheduler/core/eventsqueue/nightchanges.py
+++ b/scheduler/core/eventsqueue/nightchanges.py
@@ -53,10 +53,19 @@ class NightTimeline:
 
         all_generated = []
         t = 0
+
         for entry in reversed(entries):
             pg = entry.plan_generated
 
-            partial_plan = pg[:t] if t > 0 else pg
+            if t > 0:
+                # grab partial plan at the timeslot
+                partial_plan = pg.get_slice(stop=t)
+                # modfy reflect last visit to match starting time_slot from the visit
+                # TODO: this need to be reflected in Optimizer
+                partial_plan.visits[-1].time_slots = t - partial_plan.visits[-1].start_time_slot
+            else:
+                partial_plan = pg
+
             all_generated += [v for v in reversed(partial_plan.visits)]
             if t < entry.start_time_slots:
                 t = entry.start_time_slots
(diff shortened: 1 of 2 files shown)


sraaphorst (commented):
I'm having trouble telling when and how this code is run. Could you put a bit more detail in the PR explaining this? I just want to be confident that it doesn't interfere with a Plan such that it changes the way time accounting works for partial plans.

My understanding was that this was supposed to just be a cosmetic change in the output. The other option is to have this be in the final plans, but I don't think it should be in the intermediate plans. (I'm not sure if it is.)

The `slice` notiation is a bit confusing because of the inherent concepts of slices in Python.

The Jira task description makes sense and seems right, but I don't feel like it's quite clear what's happening here.

stroncod:
> My understanding was that this was supposed to just be a cosmetic change in the output. The other option is to have this be in the final plans, but I don't think it should be in the intermediate plans. (I'm not sure if it is.)

If the observation was modify shouldn't affect the time accounting. We could the time accounting before this and it should not affect it. 

> The slice notiation is a bit confusing because of the inherent concepts of slices in Python.
It behaves exactly as the slices in Python. I just did not wanted to override the `[]` operand.

sraaphorst:
Okay. Sounds good, and thanks for the elaboration. Can you do a quick check to make sure that `run_with_events.py` produces the same output? If so, let's merge it. I'll approve now.

sraaphorst (approved):

PR #347, opened 2023-10-31
SCHED-512: Remove process manager

stroncod:


sraaphorst:
Looks like the tests are failing here:

```
INTERNALERROR> scheduler/core/builder/builder.py:10: in <module>
INTERNALERROR>     from .blueprint import CollectorBlueprint, OptimizerBlueprint
INTERNALERROR> E   ModuleNotFoundError: No module named 'scheduler.core.blueprint'
```

sraaphorst:
Are we sure that we won't need this at all? I'm all for removing it if we don't, but I just wanted to check in.

stroncod:
> Are we sure that we won't need this at all? I'm all for removing it if we don't, but I just wanted to check in.

Is not being used and further is not going to be used as we always are going to output the final plan.

sraaphorst (approved):

PR #349, opened 2023-11-01
SCHED-486: Observations marked ON_HOLD or PHASE2 are no longer included

sraaphorst:
The only observations that are parsed from the JSON code in the OCS program parser are those that are:
* `READY`
* `ONGOING`
* `OBSERVED`
We want the latter since we reset all observations after they have been read into the `Collector`. Since this is past semester data, most (if not all) observations will be marked `OBSERVED`.

The `Selector` code now ignores `INACTIVE` and `OBSERVED` observations (we will be flipping some observations to INACTIVE based on program splitting with unnecessary tellurics, for example), and the `Selection` only contains `READY` and `ONGOING` observations.

This generates a slightly different plan as expected since certain observations that were formerly admitted but should not have been are now properly excluded.

### DIFF
--- a/scheduler/core/components/selector/__init__.py
+++ b/scheduler/core/components/selector/__init__.py
@@ -287,13 +287,12 @@ class Selector(SchedulerComponent):
             raise ValueError(f'Non-observation group {group.id} cannot be treated as observation group.')
 
         obs = group.children
-        # *** HERE ***
-        # if obs.status in {ObservationStatus.OBSERVED, ObservationStatus.INACTIVE}:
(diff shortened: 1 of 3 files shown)


sraaphorst:
This of course now causes `test_ocs_api.py` to fail. I will fix this later.

bryanmiller:
We need to keep importing OBSERVED observations initially, and then unobserve them as before. We just need to ignore obs at stages before Ready. Once we start handling ToOs we’ll need to do something with those that are On Hold.

PR #351, opened 2023-11-07
SCHED 516: Clean up

stroncod:


scheduler/core/components/optimizer/greedymax.py
```diff
@@ -530,10 +530,10 @@ def place_standards(self,
             xmean_cal = self.mean_airmass(partcal_obs.id, interval[slot_start:slot_end + 1], night_idx=night_idx)
 
             if verbose:
-                print(f'Standard {partcal_obs.id.id}')
+                print('Standard {partcal_obs.id.id}')
```
sraaphorst, line 533: Don't we need this to be an f-string? I'm not sure why the f was taken out.

scheduler/services/resource/base.py
```diff
@@ -272,7 +272,7 @@ def _process_file(f) -> None:
                 # Advance the previous row date where data was defined.
                 prev_row_date = row_date
 
-        if type(data_source) is str:
+        if isinstance(data_source, str):
```
sraaphorst, line 275: Much better!

sraaphorst (commented):
Why all the shortening of the
`from x import y, z`
imports to:
`from x import *`
?

We should try to be specific as to avoid cluttering imports through other imports.

scheduler/core/components/optimizer/greedymax.py
sraaphorst: Approving so you can go ahead and merge after you make this small change!

sraaphorst (approved):

PR #356, opened 2023-12-05
SCHED 518: Add mutations for events

stroncod:


scheduler/core/time.py
```diff
@@ -0,0 +1,21 @@
+from datetime import datetime, timedelta
```
sraaphorst, line 1: Just a pedantic note: copyright notification. 😉

scheduler/core/time.py
```diff
@@ -0,0 +1,21 @@
...
+from math import ceil
+
+from lucupy.minimodel import NightIndex
+
+
+class TimeHandler:
+
+    def __init__(self, num_nights_to_schedule, start, time_slot_length):
```
sraaphorst, line 9: Type hints? Why not make this a dataclass if you're just setting members? It doesn't look like you're changing the values as far as I can tell, so it could even be:

```python
@dataclass(frozen=True)
class TimeHandler:
    ...
```

scheduler/core/time.py
```diff
@@ -0,0 +1,21 @@
...
+        self.start = start
+        self.num_nights_to_schedule = num_nights_to_schedule
+        self.time_slot_length = time_slot_length
+
+    def dt2night_index(self, dt: datetime) -> NightIndex:
+        days = dt.day - self.start.day
+        if days > self.num_nights_to_schedule:
+            raise ValueError('Date outside of the number of night to schedule!')
```
sraaphorst, line 17: This seems like an error message that could contain more information in case it comes up... something like:

```python
max_day = self.start.day + datetime(days=self.num_nights_to_schedule - 1)
raise ValueError(f'Date {dt.strftime("%Y-%m-%d")} outside of valid range for scheduling: {self.start.strftime("%Y-%m-%d")} - {max_day.strftime("%Y-%m-%d")}')
```

I'm not quite sure what the type of `start` is.

scheduler/graphql_mid/schema.py
```diff
@@ -1,27 +1,31 @@
...
+from .inputs import CreateNewScheduleInput, UseFilesSourceInput, NewWeatherChangeInput
 from .scalars import SOrigin
-
+from ..config import config
+from ..core.time import TimeHandler
 
 # TODO: This variables need a Redis cache to work with different mutation correctly
 sources = Sources()
 event_queue = EventQueue(frozenset([NightIndex(i) for i in range(3)]), ALL_SITES)
+time_handler = TimeHandler(start=None, num_nights_to_schedule=None, time_slot_length=config.collector.time_slot_length)
```
sraaphorst, line 28: Here you set the `TimeHandler` start to `None`... won't this cause an exception when you use the `TimeHandler` methods?

scheduler/graphql_mid/schema.py
```diff
@@ -83,6 +87,32 @@ def change_origin(self, new_origin: SOrigin, mode: SchedulerModes) -> ChangeOrig
...
+    def add_weather_change(self, new_weather_change: NewWeatherChangeInput) -> EventsAddedResponse:
+        try:
+            new_cc = new_weather_change.new_CC
+            new_sb = new_weather_change.new_SB
+            new_wv = new_weather_change.new_WV
+            new_iq = new_weather_change.new_IQ
+            c = Conditions(cc=CloudCover[new_cc.name] if new_cc else None,
+                           sb=SkyBackground[new_sb.name] if new_sb else None,
+                           wv=WaterVapor[new_wv.name] if new_wv else None,
+                           iq=ImageQuality[new_iq.name] if new_iq else None)
```
sraaphorst, line 100: I'm still concerned about these being `None` if they aren't defined because the types in `Conditions` aren't `Optional`s so we should never be setting them to `None`? We should think of an alternate way to do this.

scheduler/graphql_mid/schema.py
```diff
@@ -83,6 +87,32 @@ def change_origin(self, new_origin: SOrigin, mode: SchedulerModes) -> ChangeOrig
...
+            wc = WeatherChange(event_id=uuid.uuid4(),
+                               start=new_weather_change.start,
+                               reason=new_weather_change.reason,
+                               site=new_weather_change.site,
+                               new_conditions=c)
+            event_queue.add_event(time_handler.dt2night_index(new_weather_change.start), new_weather_change.site, wc)
+        except KeyError:
+            return EventsAddedError(error='Weather Condition not valid!')
```
sraaphorst, line 108: Could we give more information about what was not valid?

scheduler/graphql_mid/schema.py
```diff
@@ -83,6 +87,32 @@ def change_origin(self, new_origin: SOrigin, mode: SchedulerModes) -> ChangeOrig
...
+        else:
+            return EventsAddedSuccess(added_event=f'Weather change added at {new_weather_change.start.timestamp()}')
+
+    @strawberry.mutation
+    def remove_weather_change(self, night: int, site: Site, event_id: uuid.UUID) -> EventRemovedResponse:
+        return EventRemovedResponse(success=event_queue.remove_event(NightIndex(night), site, event_id))
```
sraaphorst, line 114: This seems like an interesting addition and I'm curious about the use cases: when would we remove a weather change? (It's not a bad idea at all. I'm just wondering what you're thinking.)

scheduler/graphql_mid/types.py
```diff
@@ -105,6 +106,13 @@ def for_site(self, site: Site) -> 'SPlans':
             plans_per_site=[plans for plans in self.plans_per_site if plans.site == site])
 
 
+@strawberry.enum
```
sraaphorst, line 109: Just to affirm: we'll have to add event types here as we create more, right?
Could you document this process somewhere in a short blurb since we will be adding a lot more event types and I don't want to miss an essential detail?

scheduler/scripts/run_main.py
```diff
@@ -69,12 +70,18 @@ def main(*,
         night_events = collector.get_night_events(site)
         for night_idx in night_indices:
             eve_twilight_time = night_events.twilight_evening_12[night_idx].to_datetime()
-            eve_twilight = EveningTwilight(start=eve_twilight_time, reason='Evening 12° Twilight', site=site)
+            eve_twilight = EveningTwilight(event_id=uuid.uuid4(),
```
sraaphorst, line 73: If we auto-generate the `uuid` like above, then we don't have to keep initializing this field.

scheduler/graphql_mid/types.py
```diff
@@ -198,33 +206,13 @@ class SourceFileHandlerResponse:
...
-                       wv=WaterVapor[self.new_WV.name] if self.new_WV else None,
-                       iq=ImageQuality[self.new_IQ.name] if self.new_IQ else None)
-        return WeatherChange(start=self.start,
-                             reason=self.reason,
-                             new_conditions=c
-                             )
-
-
 @strawberry.type
 class EventsAddedSuccess:
```
sraaphorst, line 210: I'm thinking we might want to include the `uuid` in the `EventsAddedSuccess` so we can easily refer to an event by its `uuid`.

scheduler/scripts/run_main.py
```diff
@@ -69,12 +70,18 @@ def main(*,
...
+                                           start=eve_twilight_time,
+                                           reason='Evening 12° Twilight',
+                                           site=site)
             queue.add_event(night_idx, site, eve_twilight)
 
             # Add one time slot to the morning twilight to make sure time accounting is done for entire night.
             morn_twilight_time = night_events.twilight_morning_12[night_idx].to_datetime()
-            morn_twilight = MorningTwilight(start=morn_twilight_time, reason='Morning 12° Twilight', site=site)
+            morn_twilight = MorningTwilight(event_id=uuid.uuid4(),
```
sraaphorst, line 81: Same comment as above.

sraaphorst (commented):
Some notes for discussion.

scheduler/graphql_mid/schema.py
stroncod: To remove already created events and change the validation.

PR #358, opened 2023-12-06
SCHED-525: Fix ordering of time accounting

sraaphorst:
This repairs the ordering of time accounting in the event handling: time accounting is done before a new selection is produced and a new plan is made by the optimizer.

This also includes numerous other fixes and improvements, such as cleanup, logging, safety measures about not using possibly undefined types, and improvements on how plans are assembled.


scheduler/scripts/run_main.py
```diff
@@ -111,9 +114,10 @@ def main(*,
             selector.update_cc_and_iq(site, cc_value, iq_value)
 
         for site in collector.sites:
-            # The starting twilight for the night for the site.
+            # Initial values until the evening twilight is executed, and a plan is made.
             night_start: Optional[datetime] = None
             night_done = False
+            plans = None
```
stroncod, line 120: This confirms what Bryan was saying about the plans being done one site at it time. Even if they have access at each other in the Selection phase we do it one output at time.

scheduler/scripts/run_main.py
```diff
@@ -147,39 +152,68 @@ def main(*,
...
-                if not night_done:
+                # Calculate the time slot of the event. Note that if the night is done, it is None.
+                if night_done:
+                    event_start_time_slot = None
+                    end_timeslot_bounds = None
+                else:
                     event_start_time_slot = event.to_timeslot_idx(night_start, collector.time_slot_length.to_datetime())
+                    end_timeslot_bounds = {site: TimeslotIndex(event_start_time_slot)}
+
+                # If tbe following conditions are met:
```
stroncod, line 163: the

scheduler/scripts/run_main.py
```diff
@@ -147,39 +152,68 @@ def main(*,
...
-                                       event,
-                                       plans[0][site])
-
-                # Perform the time accounting. If the night is done, we execute it to completion.
-                end_timeslot_bounds = None if night_done else {site: TimeslotIndex(event_start_time_slot)}
-                collector.time_accounting(plans[0],
-                                          sites=frozenset({site}),
-                                          end_timeslot_bounds=end_timeslot_bounds)
+                    # Right now the optimizer generates List[Plans], a list of plans indexed by
+                    # every night in the selection. We only want the first one, which corresponds
```
stroncod, line 191: This look like a hack, we should change that to one 'Plans' or Site Plans.

scheduler/scripts/run_main.py
```diff
@@ -147,39 +152,68 @@ def main(*,
...
         # Piece together the plans for the night to get the overall plans.
+        # This is rather convoluted because of the confusing relationship between Plan, Plans, and NightlyTimeline.
+        # TODO: There appears to be a bug here. See GSCHED-517.
+        logger.info(f'Assembling plans for night index {night_idx}.')
+        night_events = {site: collector.get_night_events(site) for site in collector.sites}
+        final_plans = Plans(night_events, NightIndex(night_idx))
         for site in collector.sites:
-            plans[0][site] = night_timeline.get_final_plan(NightIndex(night_idx), site)
-        overall_plans[night_idx] = plans[0]
+            final_plans[site] = nightly_timeline.get_final_plan(NightIndex(night_idx), site)
```
stroncod, line 209: Maybe NightlyTimelime should return the final plans instead of having part of that logic outside.

stroncod (approved):

scheduler/scripts/run_main.py
sraaphorst: What GM should be doing is creating one plan per site for the given night, although right now, it creates a `List[Plan]` for all the nights for which it has data in the `Selection`. It's a bit hacky, but it's flexible: I just make sure that we only have one night in the `Selection` and GM then only prepares a `List[Plans]` with that one night.

Ultimately, I'm not sure what we are going to want: GM to still be able to produce `Plans` across multiple nights, or simplify it down to one `Plans` for the current night. The latter will, I think, require a lot of propagation through the code and offer less generalized functionality, but until those decisions are made, this is going to remain a bit hacky but functional.

PR #360, opened 2023-12-14
SCHED 527: Events from files

stroncod:


scheduler/scripts/run_main.py
```diff
@@ -71,15 +71,19 @@ def main(*,
     # The morning twilight will force time accounting to be done on the last generated plan for the night.
     for site in sites:
         night_events = collector.get_night_events(site)
-        for night_idx in night_indices:
+        curr = start.to_datetime()
+        night_idx = 0
```
sraaphorst, line 75: Discussion.

scheduler/scripts/run_main.py
```diff
@@ -104,7 +108,9 @@ def main(*,
     overall_plans: Dict[NightIndex, Plans] = {}
     nightly_timeline = NightlyTimeline()
 
-    for night_idx in sorted(night_indices):
+    curr = start.to_datetime()
+    night_idx = NightIndex(0)
+    while curr <= end.to_datetime():
```
sraaphorst, line 113: This seems like a common pattern. Is there any way we can extract it and pass in a function with the necessary details to perform each step?

scheduler/services/resource/base.py
```diff
@@ -186,6 +195,20 @@ def get_resources(self, site: Site, night_date: date) -> FrozenSet[Resource]:
...
+    def get_faults(self, site: Site, night_date: date):
+        if site not in self._sites:
+            raise ValueError(f'Request for fault for illegal site: {site.name}')
+        if date not in self._faults[site]:
+            return set()
+        return self._faults[site][night_date]
+
+    def get_eng_tasks(self, site: Site, night_date: date):
+        if site not in self._sites:
+            raise ValueError(f'Request for fault for illegal site: {site.name}')
```
sraaphorst, line 207: Change text from `fault` to `engineering tasks`.

scheduler/services/resource/base.py
```diff
@@ -575,12 +598,99 @@ def none_to_str(value) -> str:
...
+                        self._eng_task[site][just_date].add(eng_task)
+                    else:
+                        self._eng_task[site][just_date] = {eng_task}
+                else:
+                    raise ValueError(f'Pattern not found. Format error on Eng Task file at line {line_num}')
+
+    def _parse_faults_file(self, site: Site, to_file: str) -> None:
+        """Parse faults from files.
+        This is purposeful left non-private as might be used with incoming files from
+        the React app.
```
sraaphorst, line 654: Maybe we should remove the leading `_`?

scheduler/services/resource/base.py
```diff
@@ -575,12 +598,99 @@ def none_to_str(value) -> str:
...
+        """
+        # Files contains this repetitive string in each timestamp, if we need them
+        # we could add them as constants.
+        ts_clean = ' 04:00' if site is Site.GS else ' 10:00'
```
sraaphorst, line 658: Instead of doing this, can we just preprocess the files to not have this redundancy? Do a find / replace and just get rid of it?

sraaphorst (commented):
This is very good stuff, but let's discuss a few of the points here.

scheduler/core/eventsqueue/event_queue.py
```diff
@@ -35,42 +38,35 @@ def add_event(self, event: Event) -> None:
...
+        night_idx = 0
+        while curr <= end.to_datetime():
+            self._events[curr] = {site: NightEventQueue(night_idx=NightIndex(night_idx), site=site) for site in sites}
+            night_idx += 1
+            curr += timedelta(days=1)
+
         self._blockage_stack = []
 
-    def add_event(self, night_idx: NightIndex, site: Site, event: Event) -> None:
+    def add_event(self, night_date: datetime, site: Site, event: Event) -> None:
```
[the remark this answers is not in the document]
stroncod: The event queue is initialized by date, is easier in the service to grab the date than using the NightIndex.

scheduler/core/eventsqueue/events.py
```diff
@@ -73,34 +73,21 @@ class WeatherChange(Interruption):
...
-
-
 @final
-class Fault(Blockage):
+@dataclass(unsafe_hash=True)
+class Fault(Interruption):
     """
     Blockage that occurs when one or more resources experience a fault.
     """
+    id: str
```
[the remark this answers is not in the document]
stroncod: is the ID in the file, is a string with the initials `FR` is just part of the parser, I guess it can be called something else.

scheduler/core/eventsqueue/events.py
```diff
@@ -73,34 +73,21 @@ class WeatherChange(Interruption):
...
+    time_loss: timedelta
     affects: FrozenSet[Resource]
+
+
+@final
+@dataclass(unsafe_hash=True)
+class EngTask(Interruption):
+
+    end: datetime
+    time_loss: timedelta
```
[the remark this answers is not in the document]
stroncod: Yes, I didn't know how to call it so I just went with  `Interruption`

scheduler/scripts/run_main.py
stroncod: There is a lot of duplicate code between the scripts and Service.

scheduler/services/resource/base.py
stroncod: I'm wasn't sure if it was going to be needed or not.

scheduler/core/eventsqueue/event_queue.py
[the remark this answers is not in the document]
sraaphorst: @stroncod If we had functions to convert back-and-forth between `datetime` and what I'm calling "time coordinates" (`NightIndex`, `TimeslotIndex`), either local or UTC, would that make it easier? I'd just like this to be consistent.

scheduler/core/eventsqueue/event_queue.py
```diff
@@ -35,42 +38,35 @@ def add_event(self, event: Event) -> None:
...
-    def check_blockage(self, resume_event: ResumeNight) -> Blockage:
-        if self._blockage_stack and len(self._blockage_stack) == 1:
-            b = self._blockage_stack.pop()
-            b.ends(resume_event.start)
-            return b
-        raise RuntimeError('Missing blockage for ResumeNight')
+                    raise KeyError(f'Could not add event {event} for date {night_date} to site {site.name}.')
 
-    def get_night_events(self, night_idx: NightIndex, site: Site) -> Optional[NightEventQueue]:
+    def get_night_events(self, night_date: datetime, site: Site) -> Optional[NightEventQueue]:
```
sraaphorst, line 61: Like here, we are already working with `NightIndex` in the code: it would be really nice to still have this as `NightIndex` and not changed to `night_date: datetime`, which is more concerning to me because it's `datetime` (what is the time?) instead of `datetime.date`.

scheduler/core/eventsqueue/events.py
```diff
@@ -73,34 +73,21 @@ class WeatherChange(Interruption):
...
 @final
-class Fault(Blockage):
+@dataclass(unsafe_hash=True)
+class Fault(Interruption):
     """
     Blockage that occurs when one or more resources experience a fault.
     """
+    id: str
+    end: datetime
+    time_loss: timedelta
```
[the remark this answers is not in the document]
sraaphorst: This should probably be divided into two events:

1. `Fault`
2. `FaultResolved`

This would correspond to what we will see in Operations mode.

In Validation mode, we know from the data in the files:
1. When a fault occurs
2. When a fault resolves

but I think we are trying to model two concepts here into one thing, which is going to create significant technical debt for later. We could instead model each fault entry in a Validation mode file into both a `Fault` and a `FaultResolved` event linked by ID.

* The `Fault.start` would be the start time of the fault.
* The `FaultResolved.start` would be the start time at which the fault was resolved.

Then we can get rid of:
* `end`
* `timeloss`

and `timeloss` can just be calculated by:

```python
fault_resolved.start - fault.start
```

It may be misleading to call a `FaultResolved` an `Interruption`, so we might want to tweak the event hierarchy slightly, but I don't think we need to invest a lot of time there.

scheduler/core/eventsqueue/events.py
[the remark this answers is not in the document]
sraaphorst: `Interruption` works fine for the superclass name, but more what I meant is why do we repeat the fields here instead of moving them up.

If we get rid of `end` and `time_loss` as I propose above, then we can also mark the dataclass as `frozen` and that means we don't need the `unsafe_hash`, and we should try to limit the use of `unsafe_hash` as much as possible since it is inherently unsafe.

At the very least, we should mark mutable fields as not being included in the hash with:

```python
@dataclass(unsafe_hash=True)
class MyClass:
    immutable_field: type
    mutable_field: type = field(hash=False)
```

or something to that extent.

scheduler/services/resource/base.py
sraaphorst: I think we need a standardized file format, and then we take the existing files and make them match the standardized format.

We send the standard out to science staff, and use it internally if we are accepting text files for these values.

So, we:
1. Create the standard
2. Make the existing files match the standard (with a script to just read and fix them, or by fixing them by hand, but this should not be run by the Scheduler)
3. Use the standardized output files in the Scheduler

sraaphorst (changes requested):
As per PR comments from last reviews:
1. Discussion regarding switching from `NightIndex` to `datetime` for event queues.
2. `Fault` class not compatible with GPP.

PR #363, opened 2023-12-16
GSCHED-489: updated collector time accounting

bryanmiller:
Here is the next iteration of collector time accounting. It still needs some tweaking but it would be good to test what is here
- don't charge scheduling groups if an event occurs during the group visit (TODO: just do this if there is more than one observation in the group)
- observation not in a scheduling group will charge the atoms up to an interrupting event
- set unneeded partner calibrations (e.g. tellurics) to INACTIVE

Other TODO: add no_charged time category and charge interrupted observations to that.

Please review and clean up, but don't remove the commented-out print statements, I may need those for the next iteration.

### DIFF
--- a/scheduler/core/components/collector/__init__.py
+++ b/scheduler/core/components/collector/__init__.py
@@ -21,6 +21,7 @@ import numpy as np
 from scheduler.core.calculations import NightEvents, TargetInfo, TargetInfoMap, TargetInfoNightIndexMap
 from scheduler.core.components.base import SchedulerComponent
 from scheduler.core.components.nighteventsmanager import NightEventsManager
+from scheduler.core.plans import Plan, Plans
 from scheduler.core.programprovider.abstract import ProgramProvider
 from scheduler.core.plans import Plans, Visit
 from scheduler.services.resource import NightConfiguration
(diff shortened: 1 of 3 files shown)


scheduler/core/components/collector/__init__.py
```diff
@@ -503,6 +519,33 @@ def night_configurations(self,
             self.get_night_events(site).time_grid[night_idx].datetime.date() - Collector._DAY
         ) for night_idx in night_indices}
 
+    def _get_group(self, observation: Observation) -> Group:
```
sraaphorst, line 522: We will still need to figure out what we need if an `Observation` is a member of multiple levels of scheduling `Group`s.

scheduler/core/components/collector/__init__.py
```diff
@@ -503,6 +519,33 @@ def night_configurations(self,
...
+        """Return the group that an observation is a member of"""
+        outgroup = None
+
+        def _find_obs(group: Group, observation: Observation) -> bool:
+            found = False
+            for obs in group.observations():
+                if obs == observation:
+                    found = True
+                    break
+            return found
```
sraaphorst, line 532: Simplify logic here.

scheduler/core/components/collector/__init__.py
```diff
@@ -526,37 +570,122 @@ def time_accounting(self,
...
+            # print(f'\ntime_accounting')
+            # print(f'{plan.site} {end_timeslot_bound}')
+
+            group_visits = []
+            for ii, visit in enumerate(sorted(plan.visits, key=lambda v: v.start_time_slot)):
+                obs = self.get_observation(visit.obs_id)
+                group = self._get_group(obs)
+                # print(
+                #     f'{ii} {visit.obs_id.id} {group.unique_id.id} {visit.start_time_slot} {visit.start_time_slot + visit.time_slots - 1}')
+                if ii > 1 and group.is_scheduling_group() and group == group_visits[-1].group:
```
sraaphorst, line 582: I think I see what is happening here, but the logic is a bit dangerous: `group_visits` is declared empty and indexing it here could cause an `IndexError`. I think the `ii > 1` circumvents this, but we should look at something more robust.

sraaphorst (approved):
Just some comments for us to review after the code is merged.

PR #364, opened 2023-12-19
SCHED-489 cleanup

sraaphorst:
This cleans up the branch SCHED-489 by:
* Fixing some bugs (e.g. `GroupVisit` class declaration, checking method `Group.is_scheduling_group` instead of invoking it).
* Simplifying and cleaning up logic.
* Reducing `Observation` comparisons to ID comparisons instead of entire `Observation` comparisons, which compare everything (except sequences, as this functionality has been temporarily suspended).

### DIFF
--- a/scheduler/core/plans/__init__.py
+++ b/scheduler/core/plans/__init__.py
@@ -83,8 +83,7 @@ class Plan:
             # TODO: can this be done w/o a loop? convert cumm_seq to slots, and find the value that fits
             while n_slots_filled + visit_length <= len_interval and atom_end <= len(cumul_seq) - 2:
                 atom_end += 1
-                visit_length = n_slots_acq + \
-                    Plan.time2slots(self.time_slot_length, cumul_seq[atom_end])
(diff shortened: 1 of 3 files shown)


scheduler/core/plans/__init__.py
```diff
@@ -102,8 +101,9 @@ def nir_slots(self,
 
     @staticmethod
     def time2slots(time_slot_length: timedelta, time: timedelta) -> int:
```
stroncod, line 103: this is scattered around the code the same logic

stroncod (approved):

scheduler/core/plans/__init__.py
sraaphorst: I'm not quite sure what you mean... do you mean we have this logic implemented several times (in which case, we should conglomerate them)? I see a lot of calls to this method (which should probably be a function in lucupy) but I don't off the top of my head see it implemented more than once?

sraaphorst: Oh, I just did a search for `ceil`, and now I see it all over. I'll spend a bit of time cleaning that up today.

PR #365, opened 2023-12-19
SCHED-531: Move all time to num timeslots calculations to point to lucupy time2slots

sraaphorst:
As pointed out by @stroncod, the logic to convert a time quantity to number of time slots was repeated throughout the code base. This method has no inherent tie to `Plan`, where it was included before, and the logic has been moved to lucupy and all of the implementations (and other calculations in the code that calculate this explicitly) now point there.

**NOTE:** This requires an update to lucupy 0.1.59.


stroncod (approved):
Great change!

PR #368, opened 2023-12-21
SCHED-534: Methods to convert between time coordinates and datetimes

sraaphorst:
Methods on `NightEvents` to convert between `NightIndex` and `TimeslotIndex` and `datetime` objects in both directions, both using UTC and localized time.


stroncod (approved):

stroncod:
This is going to be very useful.

PR #373, opened 2023-12-28
SCHED-547: Fixing bugs in plan assembly when there are no scheduled visits.

sraaphorst:
When an event occurred and there were no `Visit`s in the plan, the code would crash due to an `IndexError` exception trying to access `partial_plan.visits[-1]`. This PR fixes this and provides clearer documentation as to what is happening.

It also fixes bugs in output when there are empty plans (i.e. providing a default value to the `max` function if a generator is empty), and more clearly indicates that there is no plan for a site for a given night when the plan is empty.

This also fixes some minor bugs and renames `try_scheduler.ipynb` to `run_validation.ipynb` for clarity.

### DIFF
--- a/demo/run_validation.ipynb
+++ b/demo/run_validation.ipynb
@@ -2,11 +2,11 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": 1,
    "metadata": {
     "ExecuteTime": {
-     "end_time": "2023-10-26T23:54:02.959513Z",
-     "start_time": "2023-10-26T23:54:02.948170Z"
+     "end_time": "2023-12-28T03:27:19.723344Z",
+     "start_time": "2023-12-28T03:27:19.713460Z"
     }
    },
    "outputs": [],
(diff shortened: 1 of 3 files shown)


scheduler/core/eventsqueue/nightchanges.py
```diff
@@ -58,11 +58,23 @@ def get_final_plan(self, night_idx: NightIndex, site: Site) -> Plan:
...
+
+                # If there was a last_visit (i.e. partial_plan.visits[-1]), then it:
+                # * started at last_visit.start_time_slot;
+                # * was scheduled for last_visit.time_slots; and thus
+                # * ended at last_visit.start_time_slot + last_visit.time_slots.
+                # If the visit was cut off, i.e:
+                #     t in [last_visit.start_time_slot, last_end_time_slot)
+                # then mark the actual number of time slots the visit was able to run.
                 # TODO: this need to be reflected in Optimizer
-                partial_plan.visits[-1].time_slots = t - partial_plan.visits[-1].start_time_slot
```
stroncod, line 65: Great catch!

stroncod (approved):

demo/run_validation.ipynb
```diff
@@ -18,14 +18,26 @@
...
     }
    },
-   "outputs": [],
+   "outputs": [
+    {
+     "ename": "ModuleNotFoundError",
+     "evalue": "No module named 'mercury'",
+     "output_type": "error",
+     "traceback": [
+      "\u001B[0;31m---------------------------------------------------------------------------\u001B[0m",
```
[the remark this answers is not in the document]
sraaphorst: It may be at this point from all the changes. I haven't run it, so this is old output. Since I updated a few things, it might work? I don't want to spend too long on it since we're going to be moving away from it.

scheduler/core/eventsqueue/nightchanges.py
sraaphorst: Thanks! I'm glad my change made no observations get filtered through at first so that we could find this now instead of later on. 😳

PR #382, opened 2024-01-09
SCHED-551: Fix source.py issues

stroncod:


scheduler/core/sources.py
```diff
@@ -101,15 +101,15 @@ def use_file(self,
...
-                    # TODO: files_input is undefined. This will be fixed later.
-                    for site in files_input.sites:
+                    for site in sites:
                         suffix = ('s' if site == Site.GS else 'n').upper()
-                        file_resource.load_files(f'GMOS{suffix}_fpu_barcode.txt',
+                        file_resource.load_files(site,
+                                                 f'GMOS{suffix}_fpu_barcode.txt',
                                                  gmos_fpu,
                                                  gmos_gratings,
                                                  calendar)
```
sraaphorst, line 110: In the most updated code, `calendar` no longer gets passed to `load_files` because it was causing the calendar to load twice (once for each site), although now I think it might make sense to "load" it twice and during each load, process the sheet for the current site... so let me look at this, fix the test cases, and get back to you.

scheduler/graphql_mid/schema.py
```diff
@@ -44,7 +44,8 @@ async def load_sources_files(self, files_input: UseFilesSourceInput) -> SourceFi
                 gmos_fpu = await files_input.gmos_fpus.read()
                 gmos_gratings = await files_input.gmos_gratings.read()
 
-                loaded = sources.use_file(service,
+                loaded = sources.use_file(files_input.sites,
+                                          service,
```
sraaphorst, line 48: I think we're going to have to include faults and eng tasks in here now.
I'm not sure about calendar, since this is a Google sheet. Let's figure out what we need to do here about that.

scheduler/core/sources.py
```diff
@@ -101,15 +101,15 @@ def use_file(self,
                 if calendar and gmos_fpu and gmos_gratings:
                     file_resource = FileResourceService()
 
-                    # TODO: files_input is undefined. This will be fixed later.
-                    for site in files_input.sites:
+                    for site in sites:
                         suffix = ('s' if site == Site.GS else 'n').upper()
-                        file_resource.load_files(f'GMOS{suffix}_fpu_barcode.txt',
+                        file_resource.load_files(site,
```
sraaphorst, line 106: So now, the calendar is loaded on a site-by-site basis according to my last PR, so this logic should work; however, we do need to pass in a faults and eng tasks file here. Check `file_resource.load_files` for the new signature.

sraaphorst (approved):
Approved, but check `load_files`.

PR #392, opened 2024-01-15
Include resource availability in visibility calculations

bryanmiller:
The visibility calculation now considers whether the required resources are available. The implementation works for validation mode, but it will probably require some refactoring for real-time mode.


sraaphorst (approved):
Good stuff. I'll go ahead and merge it.

PR #393, opened 2024-01-15
Add check for environmental variable to set the logging level.

bryanmiller:
To set logging level
1) check command line parameter
2) else look for SCHEDULER_DEBUGGING environmental variable
The latter works with Jupyter notebooks.


sraaphorst (approved):
Thanks for taking care of this!

PR #401, opened 2024-02-07
SCHED-603: Add twilights to STimelineEntryBySite

stroncod:


scheduler/scripts/run_main.py
```diff
@@ -266,6 +266,13 @@ def main(*,
                         collector.time_accounting(plans,
                                                   sites=frozenset({site}),
                                                   end_timeslot_bounds=end_timeslot_bounds)
+                        if isinstance(update.event, MorningTwilightEvent):
```
sraaphorst, line 269: Same comment as above.

scheduler/core/service/service.py
```diff
@@ -171,6 +171,12 @@ def _schedule_nights(night_indices: FrozenSet[NightIndex],
                             collector.time_accounting(plans,
                                                       sites=frozenset({site}),
                                                       end_timeslot_bounds=end_timeslot_bounds)
+                            if isinstance(update.event, MorningTwilightEvent):
```
sraaphorst, line 174: To avoid having to know about event types here, substitute:
```python
if update.done:
```

This should have the same ultimate effect without needing to muddy things with `isinstance`.

sraaphorst (changes requested):
Good, but to avoid using `isinstance`, please try the `update.done` condition and then check to see if the timeline looks correct.

sraaphorst (approved):
Thanks for making the change... I appreciate it!

PR #407, opened 2024-02-16
GSCHED-613: fix avail_resources array comparison

bryanmiller:
avail_resources needed to be compared to the other properties over the same range of sa_idx.


sraaphorst (approved):
Ah... I can't believe I missed this. Thanks for catching it.
I'll go ahead and merge and test.

PR #408, opened 2024-02-16
SCHED-614: Add file to schedule query for program selection

stroncod:


sraaphorst (approved):
👍 
Looks great. Thanks!

sraaphorst:
Ahhh... looks like this probably breaks the test cases that load program data.

PR #409, opened 2024-02-21
Import of classical programs, fix parsing of visitor instrument names

bryanmiller:
Allow import of classical programs into the collector by making them equivalent to Band 1.
Fix parsing of visitor instrument names to match the resources in the telescope calendar
Make GCAL an always available resource, which is the case. This removes a lot of warning messages but doesn't affect scheduling at this time.


bryanmiller:
I also fixed the passing of the programs list parameter and the reading of that file.

stroncod (approved):

PR #415, opened 2024-02-27
GSCHED-616: program priorities based on telescope calendar

bryanmiller:
Two ways of using the telescope calendar to adjust program score are implemented
- observation visibilities for a night are only included in the visfrac calculations if the program is allowed to be scheduled on the night (program_filter) 
- an extra program_priority term has been added to the ranker to allow programs with program_priority status to be given a scoring boost.

### DIFF
--- a/scheduler/scripts/filter_demo.py
+++ b/scheduler/scripts/filter_demo.py
@@ -2,16 +2,22 @@
 # For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
 
 import logging
+from astropy.time import Time
 
 from lucupy.observatory.abstract import ObservatoryProperties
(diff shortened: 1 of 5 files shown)


scheduler/scripts/filter_demo.py
```diff
@@ -84,6 +91,7 @@
             print(f'--- Program {pid} is rejected.')
 
     resources = frozenset({
+            resource_service.lookup_resource('Gemini South'),
```
sraaphorst, line 94: This should probably `Site.GS` or something similar insted of `'Gemin South'`, but I'll go ahead and see if I can fix this in my PR.

sraaphorst (approved):
This looks good. I'll go ahead and merge and do the rebase.

PR #416, opened 2024-02-29
SCHED-570: Hookup weather data to Env service

stroncod:


sraaphorst (commented):
Does the new data include the CC90 -> CCANY transformation?
What about other columns in the table? Is there one for WV?

Let's aim to include a bit of description in PRs and Jira tasks so that it's more evident to know what they're doing.

PR #419, opened 2024-03-05
Fix parsing GRACES sequences and calendar availability

bryanmiller:
GRACES sequences are non-standard so the acquisition steps have to be ignored.
Since GRACES is accessed via GMOS-N it does not appear in a Port column in the telescope calendar. We check the GRACES availability if GMOS-N is also available. (Note, use of the port columns and the instrument columns is redundant. In the future we could simplify this by just using the instrument columns for availability and adding the port as part of the configuration in case it is needed, eg. for AGS. This is low priority).

### DIFF
--- a/scheduler/core/programprovider/ocs/ocsprogramprovider.py
+++ b/scheduler/core/programprovider/ocs/ocsprogramprovider.py
@@ -508,19 +508,26 @@ class OcsProgramProvider(ProgramProvider):
     def _parse_instrument(data: dict) -> str:
         """Get the instrument name"""
         fpu = None
+        instrument = None
 
-        inst_key = OcsProgramProvider._AtomKeys.INSTRUMENT
-        # Visitor instrument names are in OcsProgramProvider._AtomKeys.INST_NAME
-        if OcsProgramProvider._AtomKeys.INST_NAME in data:
-            inst_key = OcsProgramProvider._AtomKeys.INST_NAME
-        instrument = data[inst_key].split(' ')[0]
+        for step in data:
+            # Ignore acquisitions
+            if step[OcsProgramProvider._AtomKeys.OBS_CLASS].upper() not in [ObservationClass.ACQ.name,
+                                                                            ObservationClass.ACQCAL.name]:
+                inst_key = OcsProgramProvider._AtomKeys.INSTRUMENT
+                # Visitor instrument names are in OcsProgramProvider._AtomKeys.INST_NAME
+                if OcsProgramProvider._AtomKeys.INST_NAME in step:
+                    inst_key = OcsProgramProvider._AtomKeys.INST_NAME
+                instrument = step[inst_key].split(' ')[0]
 
-        if instrument in OcsProgramProvider.FPU_FOR_INSTRUMENT:
-            if OcsProgramProvider.FPU_FOR_INSTRUMENT[instrument] in data:
-                fpu = data[OcsProgramProvider.FPU_FOR_INSTRUMENT[instrument]]
+                if instrument in OcsProgramProvider.FPU_FOR_INSTRUMENT:
+                    if OcsProgramProvider.FPU_FOR_INSTRUMENT[instrument] in step:
+                        fpu = step[OcsProgramProvider.FPU_FOR_INSTRUMENT[instrument]]
 
-        if instrument == 'GMOS-N' and fpu == 'IFU Left Slit (blue)':
-            instrument = 'GRACES'
+                if instrument == 'GMOS-N' and fpu == 'IFU Left Slit (blue)':
+                    instrument = 'GRACES'
+
+                break
 
         return instrument
 
@@ -542,10 +549,10 @@ class OcsProgramProvider(ProgramProvider):
             elif OcsProgramProvider.FPU_FOR_INSTRUMENT[instrument] in data:
                 fpu = data[OcsProgramProvider.FPU_FOR_INSTRUMENT[instrument]]
 
-        if OcsProgramProvider._AtomKeys.DISPERSER in data:
-            disperser = data[OcsProgramProvider._AtomKeys.DISPERSER]
-        elif instrument in ['IGRINS', 'MAROON-X']:
+        if instrument in ['IGRINS', 'MAROON-X', 'GRACES']:
             disperser = instrument
+        elif OcsProgramProvider._AtomKeys.DISPERSER in data:
+            disperser = data[OcsProgramProvider._AtomKeys.DISPERSER]
         else:
             disperser = None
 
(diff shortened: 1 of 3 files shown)


bryanmiller:
The last changes to greedymax are dependent on the following lucupy change
https://github.com/gemini-hlsw/lucupy/pull/83

scheduler/core/programprovider/ocs/ocsprogramprovider.py
```diff
@@ -508,19 +508,26 @@ def parse_nonsidereal_target(self, data: dict) -> NonsiderealTarget:
...
+                # Visitor instrument names are in OcsProgramProvider._AtomKeys.INST_NAME
+                if OcsProgramProvider._AtomKeys.INST_NAME in step:
+                    inst_key = OcsProgramProvider._AtomKeys.INST_NAME
+                instrument = step[inst_key].split(' ')[0]
 
-        if instrument in OcsProgramProvider.FPU_FOR_INSTRUMENT:
-            if OcsProgramProvider.FPU_FOR_INSTRUMENT[instrument] in data:
-                fpu = data[OcsProgramProvider.FPU_FOR_INSTRUMENT[instrument]]
+                if instrument in OcsProgramProvider.FPU_FOR_INSTRUMENT:
+                    if OcsProgramProvider.FPU_FOR_INSTRUMENT[instrument] in step:
```
sraaphorst, line 524: It seems like these if statements should be combined?
```python
if instrument in OcsProgramProvider.FPU_FOR_INSTRUMENT and OcsProgramProvider.FPU_FOR_INSTRUMENT[instrument] in step:
    fpu = step[OcsProgramProvider.FPU_FOR_INSTRUMENT[instrument]]
```
Not really a big deal, though.

sraaphorst: We could also probably simplify this down for `instrument` logic  and `fpu` logic by using the `dict` method `get(key)`, which returns `None` if `key` is not defined and the value of `key` otherwise. Just an FYI for future times this comes up.

scheduler/core/programprovider/ocs/ocsprogramprovider.py
```diff
@@ -580,11 +587,11 @@ def find_filter(filter_input: str, filter_dict: Mapping[str, float]) -> Optional
             wavelength = None
 
         # Identify GRACES
-        # if instrument == 'GRACES':
-        #     if 'DS920' in filt:
-        #         fpu = '1-fiber'
-        #     elif 'HeIIC' in filt:
-        #         fpu = '2-fiber'
+        if instrument == 'GRACES':
```
sraaphorst, line 590: Repeated strings like this littered throughout the code are magic values: we should probably just define a str at the class level, `graces = 'GRACES'`, and then use that in these computations instead of repeatedly referencing `'GRACES'` directly, but I'll make that change if we have time.

scheduler/core/programprovider/ocs/ocsprogramprovider.py
```diff
@@ -681,37 +688,40 @@ def autocorr_lag(x):
...
-                sky_p_offsets.append(p)
-                sky_q_offsets.append(q)
-            coadds.append(int(step[OcsProgramProvider._AtomKeys.COADDS])
-                          if OcsProgramProvider._AtomKeys.COADDS in step else 1)
-            exposure_times.append(step[OcsProgramProvider._AtomKeys.EXPOSURE_TIME])
-            p_offsets.append(p)
-            q_offsets.append(q)
+            # Don't consider acq steps
+            if step[OcsProgramProvider._AtomKeys.OBS_CLASS].upper() not in [ObservationClass.ACQ.name,
+                                                                            ObservationClass.ACQCAL.name]:
```
sraaphorst, line 697: It might be easier to just do:
```python
if step[OcsProgramProvider._AtomKeys.OBS_CLASS].upper() in [ObservationClass.ACQ.name, ObservationClass.ACQCAL.name]:
    continue
```

to avoid heavy nested if statements. Also, since we've reused `[ObservationClass.ACQ.name, ObservationClass.ACQCAL.name]` a couple times now, it would be better practice for us to add this at the top of the class as a static member like:

```python
acquisitions = frozenset([ObservationClass.ACQ.name, ObservationClass.ACQCAL.name])
```

or make a private method:
```python
def _check_step_is_acquisition(step: dict) -> bool:
    return step[OcsProgramProvider._AtomKeys.OBS_CLASS].upper() in [ObservationClass.ACQ.name, ObservationClass.ACQCAL.name]
```

scheduler/core/programprovider/ocs/ocsprogramprovider.py
```diff
@@ -763,109 +773,111 @@ def autocorr_lag(x):
...
-                next_atom = True
-                # logger.info('Atom for wavelength change')
-                # print(f'\t\t\t Atom for wavelength change')
-
-            # A change in exposure time or coadds is a new atom for science exposures
-            # print(f'\t\t\t {step[OcsProgramProvider._AtomKeys.OBSERVE_TYPE].upper()}')
-            if step[OcsProgramProvider._AtomKeys.OBSERVE_TYPE].upper() not in OcsProgramProvider._OBSERVE_TYPES:
-                if (prev >= 0 and observe_class.upper() == ObservationClass.SCIENCE.name and step_id > 0 and
-                        (exposure_times[step_id] != exposure_times[prev] or coadds[step_id] != coadds[prev])):
+            if observe_class.upper() not in [ObservationClass.ACQ.name, ObservationClass.ACQCAL.name]:
```
sraaphorst, line 780: Here's another place where having a function to check this would avoid repeated code.

scheduler/core/programprovider/ocs/ocsprogramprovider.py
```diff
@@ -763,109 +773,111 @@ def autocorr_lag(x):
...
-                        if step_id == 0:
-                            n_offsets += 1
-                        else:
-                            if p_offsets[step_id] != p_offsets[prev] or q_offsets[step_id] != q_offsets[prev]:
+                    # logger.info('Atom for wavelength change')
+                    # print(f'\t\t\t Atom for wavelength change')
+
+                # A change in exposure time or coadds is a new atom for science exposures
+                # print(f'\t\t\t {step[OcsProgramProvider._AtomKeys.OBSERVE_TYPE].upper()}')
+                if step[OcsProgramProvider._AtomKeys.OBSERVE_TYPE].upper() not in OcsProgramProvider._OBSERVE_TYPES:
```
sraaphorst, line 792: Something like this would work for the acquisitions, too.

scheduler/core/programprovider/ocs/ocsprogramprovider.py
```diff
@@ -763,109 +773,111 @@ def autocorr_lag(x):
...
+                    if (prev >= 0 and observe_class.upper() == ObservationClass.SCIENCE.name and step_use > 0 and
+                            (exposure_times[step_use] != exposure_times[prev] or coadds[step_use] != coadds[prev])):
```
sraaphorst, line 794: The logic here is rather confusing: I think it merits a comment for maintainability.

sraaphorst (approved):
Overall, the logic here is rather confusing and I'm not sure that the code is maintainable to someone else. Time permitting, let's see if we can simplify this (overall and by possibly extracting some of the functionality to private methods) or at least add more comments about what we are doing in the steps.

Approved now since we're on a time crunch, but I'll create a low priority Jira task to come back and revisit this when we have a chance or someone needs a cognitive shift.

PR #420, opened 2024-03-05
SCHED 633: Add weather change events 

stroncod:
Getting stuck here: 
```
 # TODO: We are getting 1-off errors sometimes by doing the weather lookups this way in the size of
 # TODO: remaining_time_slots.
 start_time = night_events.times[event_night][event_timeslot]
 end_time = night_events.times[event_night][end_time_slot]
 ```


stroncod:
I got the changemonitor working but we are getting weird offset in variables errors.

PR #423, opened 2024-03-11
SCHED-646: Remove Conda from CI

stroncod:


.github/workflows/pytest.yml
```diff
@@ -18,16 +18,16 @@ jobs:
 
       - name: Install dependencies
         run: |
-          # $CONDA is an environment variable pointing to the root of the miniconda directory
-          $CONDA/bin/conda env update --file environment.yml --name base
+            python -m pip install --upgrade pip
+            if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
```
sraaphorst, line 22: So we're not doing poetry, then?

sraaphorst (approved):

PR #429, opened 2024-03-20
Committing test code for examination.

sraaphorst:
For @stroncod to review...

### DIFF
--- a/scheduler/core/sources/origins.py
+++ b/scheduler/core/sources/origins.py
@@ -5,13 +5,18 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from enum import Enum
+from pathlib import Path
 from typing import final, NoReturn, Optional
+import pickle
 
+from definitions import ROOT_DIR
 from scheduler.services.abstract import ExternalService
 from scheduler.services.environment import OcsEnvService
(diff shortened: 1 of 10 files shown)


scheduler/core/sources/origins.py
```diff
@@ -41,11 +46,42 @@ def __str__(self):
...
+            else:
+                print(f'\tCreating env service...')
+                self.env = OcsEnvService()
+                print(f'\tPickling env service...')
+                with open(pickle_env_path, 'wb') as env_pickle:
+                    pickle.dump(self.env, env_pickle)
+
             self.is_loaded = True
+            end_timer = time.perf_counter()
+            print(f'OcsOrigin setup: {(end_timer - start_timer):.2f} seconds')
```
sraaphorst, line 84: This (minus perf timing) is what we should be doing for pickling / unpickling, I think. The logic is grouped together in one place uiniqe to OCS instead of spread around the code base.

sraaphorst: (I think it just makes sense to remove the `exists` calls and use `try`...`except...` clauses where try attempts to read the file and unpickle, and `except` creates a new instance...)

PR #433, opened 2024-03-23
SCHED-666: Pickling OCS Resource and Env service now works.

sraaphorst:
This code allows the OCS Resource and Env services to be pickled to increase the speed of the executing code.
The bulk of the speed increase is needed in the viscalcs, but this was a good test that `Resource` is now stable and working correctly.

The `Sources` do not load twice now as they did before, and logging is added to display whether the services are loaded from pickles or created.

Plans based on pickled services or freshly created services are the same, as expected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -133,6 +133,3 @@ plans.lock
 # telescope schedules for the OcsResourceService
 # Edit: include this file now as access to it online requires authorization.
 #scheduler/services/resource/data/telescope_schedules.xlsx
-
-# Pickled services
-/pickles
--- a/.gitignore
+++ b/.gitignore
@@ -120,8 +120,8 @@ maglimits_test.py
 .vscode
 .fleet
 
-# pickle
-*.pickle
+# pickle: We want Heroku to keep the pickles of OCS Env and Resource.
+#*.pickle
 
 # scala worksheet for some OCS comparison
 .metals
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -134,5 +134,5 @@ plans.lock
 # Edit: include this file now as access to it online requires authorization.
 #scheduler/services/resource/data/telescope_schedules.xlsx
 
-
-
+# Pickled services
+/pickles
```
stroncod, line 138: If there are never committed to the repo, at the moment of creating the Docker image the pickles are going to be absent and then in Heroku would basically create everything from scratch, rendering this pointless.

stroncod (changes requested):

.gitignore
sraaphorst: Fair enough. I adjusted the file.

sraaphorst:
On second thought, I disagree with the assessment. We include many files that will never be added to the repo, and it's good practice to do so:

https://github.com/gemini-hlsw/scheduler/blob/main/.gitignore

Ensuring that there can never be an accidental addition of the pickle files is important, because if they are added, they may cause problems.

Screenshot 2024-03-27 at 12 53 04 AM

sraaphorst:
Actually, never mind the previous comment: upon examination, we have `*.pickle` already in `.gitignore`, so this should be irrelevant.

stroncod (approved):

PR #437, opened 2024-04-01
SCHED-664: Connect to Redis cloud

stroncod:


scheduler/graphql_mid/schema.py
```diff
@@ -22,6 +26,8 @@
 # TODO: This variables need a Redis cache to work with different mutations correctly.
 sources = Sources()
 event_queue = EventQueue(frozenset([NightIndex(i) for i in range(3)]), ALL_SITES)
```
sraaphorst, line 28: Just a question here: should `event_queue` be set explicitly to three nights? It seems like this should be calculated, no?

sraaphorst (approved):
Approved, but could you check out the hard-coded three nights for the event queue and fix this if possible?

PR #438, opened 2024-04-02
GSCHED 650

bryanmiller:
Add a user normalized user priority term to the score.
Dependent on 
https://github.com/gemini-hlsw/lucupy/pull/94


scheduler/core/components/ranker/default.py
```diff
@@ -42,6 +42,7 @@ class RankerParameters:
     vis_power: float = 1.0
     wha_power: float = 1.0
     program_priority: float = 10.0
+    user_priority_factors: dict = {Priority.LOW: 1.0, Priority.MEDIUM: 1.25, Priority.HIGH: 1.5}.setdefault(Priority.LOW)
```
sraaphorst, line 45: For `dataclasses`, we should always specify the full type for every variable, in this case `dict[Priority, float]`, and initializing a complex value like this where you want to specify a default value should be done with `field(default_factory=...)` since doing it this way can lead to unexpected behavior. Also, the `setdefault(Priority.LOW)` is confusing because it should be the default value, and not the default key. I'm not sure what it's supposed to be doing here.

I want to make a change to this class's initialization now that I'm looking at it, so I'll do that this morning: it shouldn't affect the behavior of your changes here.

sraaphorst (approved):
Looks good except for the dictionary in `RankerParameters`, but I'll merge and fix this morning since I want to make a change anyways.

PR #442, opened 2024-04-09
GSCHED-656: Include MOS pre-imaging in score

bryanmiller:
Set Observation.preimaging from flag in ODB sequence
Add pre-imaging term to the scoring algorithm. The multiplicative preimaging_factor is a parameter. 
Make the priority_factor, from GSCHED-650, a settable parameter.


bryanmiller:
Dependent on https://github.com/gemini-hlsw/lucupy/pull/98

scheduler/core/programprovider/ocs/ocsprogramprovider.py
```diff
@@ -79,6 +79,18 @@ def read_ocs_zipfile(zip_file: str | PathLike[str], program_ids: Optional[Frozen
                 logger.info(f'Skipping program {program_id} as it is not in the list.')
 
 
+def parse_preimaging(sequence: List[dict]) -> bool:
+
+    preimaging = False
+    try:
+        if sequence[0][OcsProgramProvider._AtomKeys.PREIMAGING].upper() == 'YES':
+            preimaging = True
+    except:
```
sraaphorst, line 88: For future reference, bare exception clauses are highly recommended against.
This should probably be `KeyError`, since I think that is likely the only exception that could come up in this circumstance (unless the `sequence` might be empty, in which case, also `IndexError`.

sraaphorst (approved):
👍 I'll go ahead and merge it.

PR #447, opened 2024-04-16
GSCHED-672: add min/max altitude limits

bryanmiller:
Added min/max altitude constraints to avoid the zenith blind spot and mount low elevation/altitude limits. Scores are set to 0 outside the limits. If the defaults need to be overridden, it would be nice to be able to set the limits for only one site and not have to set both at once, as now. It would also be good to add some range checks. The minimum allowed low altitude limit is 18 degrees. The maximum high altitude limit is 90 degrees.
I also changed some plot options in greedymax for testing.


bryanmiller:
Relate PR https://github.com/gemini-hlsw/lucupy/pull/99

scheduler/core/components/ranker/default.py
```diff
@@ -29,6 +29,12 @@ def _default_score_combiner(x: npt.NDArray[float]) -> npt.NDArray[float]:
     # Note we need to use 0. or applying this function results in an array of int instead of float.
     return np.array([np.max(x)]) if 0 not in x else np.array([0.])
 
+
+# Default telescope altitude limits
+_def_alt_limits_site: Dict[Site, Dict] = {Site.GS: {'min': 18.0*u.deg, 'max': 88.0*u.deg},
+                                     Site.GN: {'min': 18.0*u.deg, 'max': 88.0*u.deg}}
+_def_alt_limits: Dict[str, float] = {'min': 18.0*u.deg, 'max': 88.0*u.deg}
+
```
sraaphorst, line 37: This is what we talked about. Instead of repeatedly using "magic" keywords of `'min'` and `'max'`, I would first create an `Enum` to represent these. This is probably best placed somewhere in lucupy since it is general and we could easily find cause to reuse this.

```python
from enum import auto, Enum

class MinMax(Enum):
    '''
    Enum typically to be used as dictionary keys when we have a dictionary
    that has both a min value and a max value key: that way we do not have
    to repeat the magic strings 'min' and 'max' in the code.

    Furthermore, it will always be the case that MIN < MAX, in case that is needed.
    '''
    MIN = auto()
    MAX = auto()
```

Then these would become:

```python
# This is just a tip to the IDE that these values should not be changed.
from typing import Dict, Final

from astropy import units as u
from astropy.coordinates import Angle
from lucupy.minimodel import Site

# If we don't wrap the angles in Angle, we get Quantity.
_def_alt_limits_site: Dict[Site, Dict[MinMax, Angle]] = {
    Site.GS: {MinMax.MIN: Angle(18.0 * u.deg), MinMax.MAX: Angle(88.0 * u.deg)},
    Site.GN: {MinMax.MIN: Angle(18.0 * u.deg), MinMax.MAX: Angle(88.0 * u.deg)}
}

# Here, you're not using a float as the value since you're multiplying by a unit.
# Again, this gives Quantity, but we probably want Angle.
_def_alt_limits: Dict[MinMax, Angle] = {
    MinMax.MIN: Angle(18.0 * u.deg),
    MinMax.MAX: Angle(88.0 * u.deg)
}

File: scheduler/core/components/ranker/default.py [hunk 2]
```diff
@@ -47,6 +53,11 @@ class RankerParameters:
     program_priority: float = 10.0
     priority_factor: float = 8.0
     preimaging_factor: float = 1.25
+    altitude_limits: Dict[str, Dict] = field(default_factory=lambda: _def_alt_limits_site)
```
sraaphorst, line 56: Do we want users of the API to be able to change `altitude_limits`?

If not, then we do this. Note the change in the parameters, which matches the definition above that you use as the default value. (The type hints in dataclasses are really important in Python.)

```python
altitude_limits: Dict[Site, Dict[MinMax, Angle]]  = field(default_factory=lambda: _def_alt_limits_size)
```

If we don't want them to be able to initialize this, we do the same thing, but we change the call to `field` to be:

```python
# Everything else stays the same.
... = field(init=False, ...)
```

Then instead of doing lookups with strings `'min'` and `'max'`, which is inadvisable (you could run into case issues if you forget, and it's just a bad idea to have identical strings littered through the code in general), you'll look up with `MinMax.MIN` and `MinMax.MAX`.

sraaphorst: Oh, I should give you an example of why we need `field` with `default` (in the case of primitives) and `default_factory` (in the case of more complex types).

Both ChatGPT and I are having trouble finding a use case where `field(..., default=...)` is really necessary outside of exceptional circumstances, so I wouldn't worry about it unless you want to apply other `field` keywords to the field, e.g:

```python
my_field: int = field(init=False, repr=False, default=5)
```

`default_factory` is much more important. To try to give you an idea why, it's because Python handles default complex data structures in a way that seems very unpredictable to most people. Here's some example code:

This is kind of a contrived example, but say you have a simple function to take an item and a list and you want to add the item to the list, and you pass in an empty list as the default:
```python
def add_to_list(item, my_list=[]):
    my_list.append(item)
    return my_list
```

Then, if you do:
```python
my_list = add_to_list('apple') # Expected: ['apple']
my_list2 = add_to_list('banana', my_list) # Expected: ['apple', 'banana']

# In this next case, you'd expect a list just containing cantaloupe due to the empty list default parameter:
my_list3 = add_to_list('cantaloupe')

# Instead, you get that my_list3 is actually ['apple', 'banana', 'cantaloupe']
```

This hopefully gives you a hint why you need to create new objects instead of using defaults, which is what `default_factory` does.

So in the case of your code above, to avoid risking changing the `_def_alt_limits_site`, you use the notation:

```python
... = field(default_factory=lambda: _def_alt_limits_site)
```
instead. Python calls the lambda expression and then makes a clean copy of `_def_alt_limits_site` to use for the dataclass.

This could be even worse if you wanted to, say, build up student grades:

```python
@dataclass
class Grades:
    grades: List[float] = []

    def add_grades(self, grade_batch: List[float]):
        self.grades.extend(grade_batch)
```

Python won't even let you assign `[]` to grades, and gives:

```python
ValueError: mutable default <class 'list'> for field grades is not allowed: use default_factory
```

sraaphorst (approved):
I'm not sure where `'max'` and `'min'` in the dictionaries come up after reading that.
I'm going to approve this, merge, and make the recommended changes. I hope that is alright! Let me know if you have concerns or if my explanation is unclear.

scheduler/core/components/ranker/default.py
bryanmiller: Thanks for the suggestions. I committed new code that implements these. The parameters for the sites can now be given separately and I added error checking. This also requires 
https://github.com/gemini-hlsw/lucupy/pull/100

sraaphorst (approved):
This looks excellent. Thanks for making the changes!

PR #450, opened 2024-04-25
SCHED-681: Implement Redis for vis calcs

stroncod:


scheduler/core/calculations/targetinfo.py
```diff
@@ -64,6 +65,64 @@ class TargetInfo:
...
+                                         dec=np.array(ti_dict['coord']['dec']) * u.deg,
+                                         frame=ti_dict['coord']['frame']),
+                          alt=Angle(np.array(ti_dict['alt']['value']), unit=ti_dict['alt']['unit']),
+                          az=Angle(np.array(ti_dict['az']['value']), unit=ti_dict['az']['unit']),
+                          par_ang=Angle(np.array(ti_dict['par_ang']['value']), unit=ti_dict['par_ang']['unit']),
+                          hourangle=Angle(np.array(ti_dict['hourangle']['value']), unit=ti_dict['hourangle']['unit']),
+                          airmass=np.array(ti_dict['airmass']),
+                          sky_brightness=np.array(ti_dict['sky_brightness']),
+                          visibility_slot_idx=np.array(ti_dict['visibility_slot_idx']),
+                          visibility_slot_filter=np.array(ti_dict['visibility_slot_filter']),
```
sraaphorst, line 80: For these two, we might not need them both: I think if we use the `NightEvents`, we can just figure out one from the other? If it's not going to take up too much space, though, we may as well include them both in the cache.

scheduler/core/calculations/targetinfo.py
```diff
@@ -64,6 +65,64 @@ class TargetInfo:
...
+                          visibility_time=TimeDelta(ti_dict['visibility_time']['value'],
+                                                    format=ti_dict['visibility_time']['format']),
+                          rem_visibility_time=TimeDelta(ti_dict['rem_visibility_time']['value'],
+                                                        format=ti_dict['rem_visibility_time']['format']),
+                          rem_visibility_frac=ti_dict['rem_visibility_frac']
+                          )
```
sraaphorst, line 86: These ones, too... I think we can figure out `rem_visibility_frac` from `visibility_time` and `rem_visibility_time`, but if we can store it all, there's no reason not to.

One thing that would be useful is to go through the code and see what fields aren't being used and remove them from dataclasses in `TargetInfo`, `Selection`, `NightEvents`, `NightConfiguration`, etc.

scheduler/core/components/collector/collector.py
```diff
@@ -301,138 +303,152 @@ def _calculate_target_info(self,
...
-                                                                            targ_prop[sa_idx] <= elev_max))))
-            )[0]
-
-            # Apply timing window constraints.
-            # We always have at least one timing window. If one was not given, the program length will be used.
-            for tw in timing_windows:
-                tw_idx = np.where(
-                    np.logical_and(night_events.times[night_idx][sa_idx[c_idx]] >= tw[0],
-                                   night_events.times[night_idx][sa_idx[c_idx]] <= tw[1])
+                # Calculate the ra and dec for each target.
```
sraaphorst, line 315: GitHub says the code starting here has changed considerably, but it's going to take awhile to manually go through and see what changed given how much is indicated. Could you give a brief summary or what the changes are in this PR?

scheduler/scripts/run_main.py
```diff
@@ -31,8 +31,8 @@
 
 def main(*,
          verbose: bool = False,
-         start: Optional[Time] = Time("2018-10-01 08:00:00", format='iso', scale='utc'),
-         end: Optional[Time] = Time("2018-10-03 08:00:00", format='iso', scale='utc'),
+         start: Optional[Time] = Time("2018-08-01 08:00:00", format='iso', scale='utc'),
+         end: Optional[Time] = Time("2018-08-03 08:00:00", format='iso', scale='utc'),
```
sraaphorst, line 35: Just curious why the change in dates here.

Also, the dates are kind of screwy: the original `2018-10-01` and `2018-10-03` actually schedules nights `2018-09-30` to `2018-10-02`, so if we do need the beginning of the semester, we should probably be doing `2018-08-02` rather than `2018-08-01` (which will probably calculate `2018-07-31`).

sraaphorst (commented):
This looks good... let's talk about it some more since it's a draft. We can discuss in the tech meeting or later.

scheduler/services/visibility/calculator.py
```diff
@@ -0,0 +1,221 @@
...
+@immutable
+@dataclass(frozen=True)
+class TargetVisibility:
+    visibility_slot_idx: npt.NDArray[int]
+    visibility_time: TimeDelta
+    rem_visibility_time: TimeDelta
+    rem_visibility_frac: float
+
+
+def calculate_target_snapshot(night_idx,
```
sraaphorst, line 32: This all looks great. Do we know the types here? Can we add them as type hints?

scheduler/services/visibility/calculator.py
```diff
@@ -0,0 +1,221 @@
...
+                          alt=alt,
+                          az=az,
+                          par_ang=par_ang,
+                          hourangle=hourangle,
+                          airmass=airmass,
+                          target_sb=targ_sb,
+                          sky_brightness=sb)
+
+
+def calculate_target_visibility(obs,
```
sraaphorst, line 99: Same here: type hints?

sraaphorst (approved):
This all looks really good and I'm impressed and happy that you got it to work before I left... I'd recommend adding some type hints and docstrings, but that's just a nitpick. Great job!

sraaphorst:
Oh, yeah: probably best to squash and merge when you do merge it instead of merging all these commits.

PR #454, opened 2024-05-13
GSCHED-672: Nonsidereal target support

sraaphorst:
This is a (currently) experimental PR to support nonsidereal targets. It requires this PR in order to run:
https://github.com/gemini-hlsw/lucupy/pull/103
along with an update to lucupy.

There are merge conflicts between this and the current branch that need to be resolved.

This code should be time-aware, i.e. the data retrieved for a `NonsiderealTarget` coordinates for a `Site` for a given night should be from sunrise to sunset (@bryanmiller said this would be needed at some point), and the code in the `Collector` should also trim the data before the eventing twilight and after the morning twilight.

It also adds some programs with nonsidereal targets to the list of program IDs to use so that the nonsidereal code will test by default when run.

I am making this a draft PR because:
1. It will not merge with changes made in the last week;
2. It requires the lucupy merge and update as given above; and
3. It should be tested more after being updated to be fit to merge.

EDIT: I meant to make this a draft PR, but I accidentally clicked on making it a full PR.
If there are any significant problems, let me know in the comments here and I will check this this week.


sraaphorst:
It looks like the test cases are not running? This should break test cases due to the change from `MAJOR_BODY` to `MAJORBODY` in lucupy?

stroncod (approved):

bryanmiller (approved):
This seems ok, let's resolve conflicts and merge so that we can test more thoroughly

PR #455, opened 2024-05-17
Gsched 627: bug fixes

bryanmiller:
I added various missing class and variable definitions so that it would run without crashing. More tests are needed that the nonsidereal calculations are correct.

### DIFF
--- a/scheduler/core/calculations/targetinfo.py
+++ b/scheduler/core/calculations/targetinfo.py
@@ -54,6 +54,7 @@ class TargetInfo:
     airmass: npt.NDArray[float]
     sky_brightness: npt.NDArray[SkyBackground]
     visibility_slot_idx: npt.NDArray[int]
+    visibility_slot_filter: npt.NDArray[float]
     visibility_time: TimeDelta
     rem_visibility_time: TimeDelta
     rem_visibility_frac: float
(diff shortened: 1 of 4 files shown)


bryanmiller:
Also, we need to merge the new program json files that go with this.

bryanmiller:
Don't test with program GN-2018B-Q-130 at the moment. The target A/2018 W1 is now called C/2018 W1 and is a comet, so the program needs to be edited,

scheduler/core/calculations/targetinfo.py
```diff
@@ -54,6 +54,7 @@ class TargetInfo:
     airmass: npt.NDArray[float]
     sky_brightness: npt.NDArray[SkyBackground]
     visibility_slot_idx: npt.NDArray[int]
+    visibility_slot_filter: npt.NDArray[float]
```
stroncod, line 57: Why the filter needs to be added again? as far as I can tell it was not being used

stroncod (approved):

scheduler/core/calculations/targetinfo.py
bryanmiller: I was getting an error in the collector if this wasn't defined.

PR #457, opened 2024-05-27
GSCHED-622: partner block handling

bryanmiller:
This updates the filter rules for partner blocks so that programs from limited term partners, e.g. Korea, can only be selected on partner nights. New ocs resource pickles files are created and should be included.
This also fixes a bug with setting program start_date and end_date for A semester programs.
Some new Horizons ephemeris files are included since new nights were used for testing.


bryanmiller:
Note, these new rules affect the visibility calculations so the redis cache will need to be rebuilt.

bryanmiller:
Also, while testing this I found that some old astropy settings were causing the timezone for Hawaii to be set wrong. I fixed this by deleting ~/.astropy/cache and then letting astropy remake the cache. I'm currently using astropy 6.1 and that has the correct Hawaii timezone.

stroncod (approved):

PR #475, opened 2024-07-17
Improve timing

dngomez:
In nightevents, time object was calculated using pandas instead of transforming from Time astropy object.
Changed visibility data format to store data ranges in redis instead of big lists, this change may break things if redis cache is not removed, since it is not compatible with the old format.


dngomez:
Tests are failing exactly because redis cache

stroncod:
Hey @dngomez could you create a task in JIRA so we can follow the format. Basically the name of the branch is the name of the JIRA task. Commits comments also has to follow the name of the JIRA task. Both the test and the script also has to run so the branch can be merged.

dngomez:
Sure, please add me to the JIRA project and I will create a ticket.
Maybe 2, to separate the changes.

PR #476, opened 2024-07-18
Improve timing in collector nightevents

dngomez:
Change nightevents utc and local time lists to be created from scratch using pandas

### DIFF
--- a/scheduler/core/calculations/nightevents.py
+++ b/scheduler/core/calculations/nightevents.py
@@ -10,13 +10,16 @@ from zoneinfo import ZoneInfo
 import astropy.units as u
 import numpy as np
 import numpy.typing as npt
+import pandas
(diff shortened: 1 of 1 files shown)


stroncod:
is this faster somehow ? I don't understand the reason of the change, can you elaborate a bit?

dngomez:
Exactly, it is faster, the Time translation from astropy to datetime (to_datetime with an specific ZoneInfo) objects takes a couple of minutes and creating  the events from scratch using pandas to have the same datetime objects takes just a few seconds.

Computing the time.time() I get.
The current code 112.71s
With the change 4.70s

stroncod (approved):

stroncod:
Great, thanks for the explanation. Approved!

PR #481, opened 2024-07-29
GSCHED-716: initial plans using GPP

bryanmiller:
Bug fixes and updates to GppProgramProvider, selector, ranker, and GreedyMax to allow the creation of plans with validation mode capabilities (consecutive AND groups only).


stroncod (approved):

scheduler/core/components/selector/selector.py
dngomez, line 1: Why print statements instead of logger.debug or something like that?

dngomez (approved):
LGTM!

PR #492, opened 2024-08-01
Modified logs to use debug and info levels

dngomez:


.github/workflows/run.yml
```diff
@@ -31,4 +31,4 @@
             REDISCLOUD_URL: ${{ secrets.REDISCLOUD_URL }}
             GPP_URL: ${{ secrets.GPP_URL }}
             GPP_KEY: ${{ secrets.GPP_KEY }}
-          run: python scheduler/scripts/run_greedymax.py
+          run: python scheduler/scripts/run_greedymax.py ERROR
```
stroncod, line 34: Even if these changes don't make so much improve in run time it would be nice to manage these more dynamically for the whole system (both the script and the service itself). Could you make it that the level is set from a env variable ? that could allows us to modify the levels without deploying. Probably through or config.yml file.

stroncod (changes requested):

.github/workflows/run.yml
dngomez: Sure, I will!

definitions.py
```diff
@@ -11,22 +11,26 @@
...
 
-# Default is full logging.
-DEFAULT_LOGGING_LEVEL = LoggingLevels.INFO
+# Default logging from LOG_LEVEL env variable otherwise INFO
+try:
+    LOG_LEVEL = os.environ['LOG_LEVEL'].upper()
+except KeyError:
+    LOG_LEVEL = 'INFO'
+
+DEFAULT_LOGGING_LEVEL = LoggingLevels[LOG_LEVEL]
```
dngomez, line 28: LOG_LEVEL environment variable now can be set to one of the possible values listed in LoggingLevels class `INFO`, `DEBUG`, `WARNING`, `ERROR`, `CRITICAL` or `FATAL`.
If LOG_LEVEL env var is not set, `INFO` will be used as default.

stroncod (approved):

stroncod:
Looks good! Thanks for the changes

PR #493, opened 2024-08-13
Gsched 730 add requested parameters to logging for each scheduler run

dngomez:


scheduler/core/components/ranker/default.py
```diff
@@ -81,6 +81,37 @@ def __post_init__(self):
             if self.altitude_limits[site][MinMax.MAX] > Angle(90.0*u.deg):
                 raise ValueError(f'The maximum altitude limit for {site.name} must be 90 degrees or less.')
 
+    def altitude_limits_to_str(self) -> str:
```
stroncod, line 84: Lets move this to `scheduler\core\output`

scheduler/graphql_mid/schema.py
```diff
@@ -114,6 +117,8 @@ async def test_sub_query(self, schedule_id: str, new_schedule_input: CreateNewSc
                                      new_schedule_input.semester_visibility,
                                      new_schedule_input.num_nights_to_schedule)
 
+        _logger.info(f"Run ID: {schedule_id}\n{params}")
```
stroncod, line 120: Nice! now it should be easier to track the request.

scheduler/core/components/ranker/default.py
stroncod, line 84: Lets move this to `scheduler/core/output`

stroncod (changes requested):
Looks good, just a small change. 
PS: Next time just use GSCHED-703 for the name of branch. JIRA can still link it and the branch names don't get too large.

scheduler/core/components/ranker/default.py
dngomez: Sure, I wasn't aware of the use of `scheduler/core/output`

dngomez: If I move this one to `scheduler/core/output`, I got an `ImportError`. There are circular imports.

In the Ranker default I am importing output -> output imports Collector -> Collector imports NightEvents -> NightEvents imports selection.py -> selection.py imports Ranker

dngomez: The import  can be done inside the __str__(self) method, but it won't be done yet, since NightEvent should be checked and maybe cleaned in the future

dngomez: Modified method to be private

stroncod (approved):
All good!

PR #494, opened 2024-08-14
SCHED 692: Add night conditions 

stroncod:
- Add night conditions to Plan to show the conditions used when the plan was created
- Adds initial condition files to OCSEnvService and use that data instead of the original calculation.


dngomez (approved):
LGTM

PR #495, opened 2024-08-14
SCHED 475: Add required conditions to each visit

stroncod:
When showing the visits now the required conditions for the observation are included in the response.


dngomez (approved):
Pretty much the same than SCHED 692 PR.

LGTM!

PR #497, opened 2024-08-21
SCHED-724: Separate time losses by type

stroncod:


stroncod:
This would make a change in the subscription changing the field `timeloss` to `timeLoss` in the `NightStats` to take in consideration in the UI

dngomez (approved):
It seems the UI should receive a JSON formated string instead of a common string.
I will change the subscription to match the new `timeLoss` key.
LGTM!

scheduler/core/statscalculator/statscalculator.py
```diff
@@ -42,12 +50,43 @@ def calculate_timeline_stats(timeline: NightlyTimeline,
         programs = {}
         for night_idx in nights:
             for site in sites:
+                time_losses = {StatCalculator._FAULT_KEY: 0,
+                               StatCalculator._WEATHER_KEY: 0,
+                               StatCalculator._UNSCHEDULE_KEY: 0}
```
dngomez, line 55: `time_losses` values are defined as an int `0`

scheduler/core/statscalculator/statscalculator.py
```diff
@@ -42,12 +50,43 @@ def calculate_timeline_stats(timeline: NightlyTimeline,
...
                     if entry.plan_generated is None:
                         continue
 
                     plan = entry.plan_generated  # Update last plan
+
+                    if isinstance(entry.event, InterruptionResolutionEvent):
+                        if isinstance(entry.event, FaultResolutionEvent):
+                            time_losses[StatCalculator._FAULT_KEY] += int(entry.event.time_loss.total_seconds()/60)
+                        elif isinstance(entry.event, WeatherClosureResolutionEvent):
+                            time_losses[StatCalculator._WEATHER_KEY] += int(entry.event.time_loss.total_seconds()/60)
```
dngomez, line 84: Like it is being used here.

scheduler/graphql_mid/types.py
```diff
@@ -34,7 +34,8 @@ class SNightStats:
     def from_computed_night_stats(ns: NightStats) -> 'SNightStats':
         cf = json.dumps(ns.completion_fraction)
         pc = json.dumps(ns.program_completion)
-        return SNightStats(timeloss=ns.time_loss,
+        tl = json.dumps(ns.time_loss)
+        return SNightStats(time_loss=tl,
                            plan_score=ns.plan_score,
                            n_toos=ns.n_toos,
                            completion_fraction=cf,
```
dngomez, line 41: Other suggestion.
These `JSON` objects, can be send exactly as they are coming in the `ns`, graphql will handle them correctly and the UI will receive an object instead of a string that needs to be parsed into an object.
This method would be
```
def from_computed_night_stats(ns: NightStats) -> 'SNightStats':
    return SNightStats(time_loss=ns.time_loss,
                       plan_score=ns.plan_score,
                       n_toos=ns.n_toos,
                       completion_fraction=ns.completion_fraction,
                       program_completion=ns.program_completion)
```

PR #500, opened 2024-08-26
Send server error to the UI

dngomez:
Changed subscription schema to send errors to UI if schedule calculation crashes

### DIFF
--- a/scheduler/graphql_mid/types.py
+++ b/scheduler/graphql_mid/types.py
@@ -189,8 +189,9 @@ class SNightTimelines:
 
 @strawberry.type
 class NewNightPlans:
-    night_plans: SNightTimelines
-    plans_summary: JSON
+    night_plans: Optional[SNightTimelines]
+    plans_summary: Optional[JSON]
+    error: Optional[str]
 
 
 @strawberry.type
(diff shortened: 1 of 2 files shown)


scheduler/graphql_mid/types.py
```diff
@@ -189,8 +189,9 @@ def from_computed_timelines(timeline: NightlyTimeline) -> 'SNightTimelines':
 
 @strawberry.type
 class NewNightPlans:
-    night_plans: SNightTimelines
-    plans_summary: JSON
+    night_plans: Optional[SNightTimelines]
+    plans_summary: Optional[JSON]
+    error: Optional[str]
```
stroncod, line 194: According to these: https://strawberry.rocks/docs/guides/errors
Strawberry catches the exception and show it in a message error. In the other PR you are parsing this new field but it sounds that we can catch the error without adding this to the return type. Could you expand more on this, why the filed is necessary vs using the included error parser?

Also it looks like the recommended implementation is using type unions
```
# Create a Union type to represent the 2 results from the mutation
Response = Annotated[
    Union[RegisterUserSuccess, UsernameAlreadyExistsError],
    strawberry.union("RegisterUserResponse"),
]
```
So the NewNightPlans should be union with other specific error types.

stroncod:
There is a failing test that is working half way: we are not using the query `schedule` method but when the fields/returns are changed in the subscription it shows up as well in there as is using the same types. Let's fix this so the CI does not show error but I'll add a task to update the testing for the graphql middleware.

scheduler/graphql_mid/types.py
dngomez: You are right, I was not receiving the error as I expected and that's why I decided to modify the schema to actually send the error, but that's not the case. The error is received in a complete different object in the client. I will throw away this PR and re-implement it using the default strawberry generated errors.

Thank you for the feedback.

PR #509, opened 2024-09-19
Fix visibility_time format to match its value (both in seconds).

MarkEverett:
Otherwise TimeDelta assumes a default format of 'jd', which will not match the value and visibility fraction factor in observation scores will be wrong.

This fix has not been tested because the calculated values have been used to fill the redis database, which I can't alter.  My understanding is that the redis database must be refilled using this fix.  Then we may have the observation scores calculated correctly.


stroncod:
Hey Mark! I tried the fix and it looks like the values are bigger now. Can you update the branch so we can merge it? I'll run the redis values after that.

MarkEverett:
Hi Sergio, I'm not certain if I did what you needed.  If you need more let me know (I'm not that familiar with the workflow I should be doing).

stroncod (approved):
All good!

PR #517, opened 2024-10-16
SCHED-744: Add new time losses values

stroncod:


scheduler/core/statscalculator/statscalculator.py
```diff
@@ -49,44 +49,56 @@ def calculate_timeline_stats(timeline: NightlyTimeline,
...
                 for entry in timeline.timeline[night_idx][site]:
                     if isinstance(entry.event, InterruptionEvent):
                         interruptions.append(entry.event)
                     elif isinstance(entry.event, InterruptionResolutionEvent):
-                        interruptions.pop()
+                        interruptions.pop()  # remove reported interruption and register the time loss
+                        if isinstance(entry.event, FaultResolutionEvent):
+                            timeline_time_losses[StatCalculator._FAULT_KEY] += int(entry.event.time_loss.total_seconds()/60)
+                        elif isinstance(entry.event, WeatherClosureResolutionEvent):
+                            timeline_time_losses[StatCalculator._WEATHER_KEY] += int(entry.event.time_loss.total_seconds()/60)
```
dngomez, line 68: I am not sure if I understand correctly this logic.
If the event is an Interruption it is added to the interruptions array. If the event is an InterruptionResolution, the last interruption event is removed from the array. This means, always the Interrupt and Resolution events will be consecutive or there are cases where an InterruptionResolution is associated to a previous Interruption?

scheduler/core/statscalculator/statscalculator.py
```diff
@@ -49,44 +49,56 @@ def calculate_timeline_stats(timeline: NightlyTimeline,
...
+                    print('event never finished:', e.description)
                     if isinstance(e, FaultEvent):
                         time_loss = timeline.timeline[night_idx][site][-1].event.time - e.time
-                        time_losses[StatCalculator._FAULT_KEY] += time_loss.total_seconds() / 60
+                        timeline_time_losses[StatCalculator._FAULT_KEY] += int(time_loss.total_seconds() / 60)
 
                     elif isinstance(e, WeatherClosureEvent):
                         time_loss = timeline.timeline[night_idx][site][-1].event.time - e.time
-                        time_losses[StatCalculator._WEATHER_KEY] += time_loss.total_seconds() / 60
+                        timeline_time_losses[StatCalculator._WEATHER_KEY] += int(time_loss.total_seconds() / 60)
```
dngomez, line 79: I guess this list may have multiple elements and it seems that time_loss is being calculated as the time between the interruption and the last night event.
This means, if you have 2 unresolved interruptions, for example, one of type FaultEvent and the other of WeatherClosureEvent, both happening at the beginning of the night, the time_loss associated to the FaultEvent will be the entire night time, and the same will happen for the WeatherClosureEvent, isn't this wrong?

By the way, I think this loop can be join with the previous one.

scheduler/core/statscalculator/statscalculator.py
stroncod: They are consecutive. If a fault happens other events might occur ( weather change but not an interruption) but if is weather closure everything is closed so nothing happens in between.

scheduler/core/statscalculator/statscalculator.py
stroncod: As the same above if the weather closure is happening we will not have a fault but if the fault happens is for the rest of the night. Is not possible to have both at the same time. what it could happen: is fault - faultResolve - WeatherClose. Then the fault is calculated above and the weather closed is for the rest of the night.

dngomez (approved):
LGTM!

PR #518, opened 2024-10-21
Changed program_file to programs_list

dngomez:


scheduler/core/programprovider/ocs/ocsprogramprovider.py
```diff
@@ -44,20 +44,15 @@
...
 
 
-def ocs_program_data(program_list: Optional[bytes] = None) -> Iterable[dict]:
+def ocs_program_data(program_list: Optional[List[str]] = None) -> Iterable[dict]:
     try:
         # Try to read the file and create a frozenset from its lines
         if program_list:
-            list_file = program_list
+            id_frozenset = frozenset(p.strip() for p in program_list if p.strip() and p.strip()[0] != '#')
+            print("ID FROZENSET")
```
stroncod, line 52: leftover prints

scheduler/core/programprovider/ocs/ocsprogramprovider.py
```diff
@@ -44,20 +44,15 @@
 DEFAULT_PROGRAM_ID_PATH = Path(ROOT_DIR) / 'scheduler' / 'data' / 'program_ids.txt'
 
 
-def ocs_program_data(program_list: Optional[bytes] = None) -> Iterable[dict]:
+def ocs_program_data(program_list: Optional[List[str]] = None) -> Iterable[dict]:
```
stroncod, line 47: Are we just passing a list of strings,  would this not break the file using in other stuff like greedymax script and fill_redis script?

stroncod (changes requested):

scheduler/core/programprovider/ocs/ocsprogramprovider.py
dngomez: Yes, you are right, it will break the other scripts, let me see what I can do to have both input types in the ocs_program_data.

stroncod (approved):

PR #521, opened 2024-11-04
SCHED-737: New test for graphql middleware

stroncod:


tests/unit/scheduler/graphql_mid/test_graphql_subs.py
```diff
@@ -0,0 +1,87 @@
...
+    result = await schema.execute(query)
+
+    async for result in sub_response:
+        # Check return without errors
+        assert not result.errors, 'Subscription returned with errors'
+        # Check the correct number of nights.
+        print([result.data["queueSchedule"]])
+        n_nights = len(result.data["queueSchedule"]["nightPlans"]["nightTimeline"])
+        print('Result night timeline: ', result.data["queueSchedule"]["nightPlans"]["nightTimeline"])
+        assert n_nights == 1, f'Number of nights must be 3, but got {n_nights}'
```
dngomez, line 73: Should be 3 instead of 1 in the assertion.
`assert n_nights == 3, f'Number of nights must be 3, but got {n_nights}'`

tests/unit/scheduler/graphql_mid/test_graphql_queries.py
dngomez, line 1: Maybe a test to check if the request was added to the queue would be nice to have, something like

```
@pytest.mark.asyncio
async def test_schedule_query_is_in_queue(set_observatory_properties):
    query = """
        query Schedule {
            schedule(scheduleId: "1", 
                     newScheduleInput: {startTime: "2018-10-01 08:00:00",
                                        endTime: "2018-10-03 08:00:00"
                                        sites: "GN", 
                                        mode: VALIDATION,
                                        semesterVisibility:false, 
                                        numNightsToSchedule:1})
        }
    """

    result = await schema.execute(query)
    assert "schedule" in result.data
    assert result.data["schedule"] is not None
    assert result.data["schedule"].startswith("Plan is on the queue!")
```

stroncod: There is not other response so if we catch a response is going to be the "Plan is on the queue message" if not we catch an error

tests/unit/scheduler/graphql_mid/test_graphql_subs.py
stroncod: Thanks! Have to change the message!

dngomez (approved):
LGTM!

PR #525, opened 2024-11-12
SCHED-757: Removing EveningTwilight when Closed Dome is happening

stroncod:


scheduler/engine/engine.py
```diff
@@ -26,6 +26,8 @@
     'Engine'
 ]
 
+from ..core.output import print_plans
```
dngomez, line 29: Is this import being used?

scheduler/services/resource/file_based_resource_service.py
```diff
@@ -494,7 +494,7 @@ def _load_time_loss(self,
                         # print(f'{local_date_str} {astropy_time} {eve_twi} {morn_twi} {site.timezone}')
 
                         if start_time is None:
-                            start_time = eve_twi.replace(second=0, microsecond=0).time()
+                            start_time = (eve_twi - timedelta(seconds=10)).time()
```
dngomez, line 497: Is this doing the trick to remove EveningTwilight?
Could you please explain the idea?

dngomez (approved):

scheduler/services/resource/file_based_resource_service.py
stroncod: It doesn't remove the event. Basically moves the Closing event to be before the Twilight and makes the telescope blocked in the event loop.

PR #529, opened 2024-12-03
Enabled with_redis functionality

dngomez:
`visibility_calculator` can be called with parameters `with_redis` and `programs_ids` if desired, otherwise everything will work as it was before this change.

### DIFF
--- a/scheduler/services/redis_client/__init__.py
+++ b/scheduler/services/redis_client/__init__.py
@@ -1,1 +1,2 @@
 from .redis_client import redis_client
+from .redis_client import RedisClient
--- a/scheduler/services/visibility/calculator.py
+++ b/scheduler/services/visibility/calculator.py
@@ -1,4 +1,3 @@
-import json
 from dataclasses import dataclass
 from typing import final, Dict, List, Any, Optional, ClassVar, FrozenSet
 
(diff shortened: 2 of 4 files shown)


scheduler/services/visibility/calculator.py
```diff
@@ -169,8 +167,17 @@ async def calculate(self) -> None:
                     _logger.info(f'Visibility calcs for {semester} from Redis retrieved.')
 
         else:
-            # fill the table manually. see fill_redis code.
-            raise NotImplementedError('Manual population not yet available')
+            from scheduler.scripts.fill_redis import compute_visibility, transform_to_json
```
stroncod, line 170: The problem with this is that we are creating two collectors, one is run in the script so we can parsed the observations and the other is the main one. I think is not a bad compromise but in this case than lets move both compute_visibility to the visibility calculator service and then transform_to_json to the redis client. Calling from the script does not look clean, the script should be that a one time interaction that ends not a module itself.

scheduler/services/redis_client/__init__.py
```diff
@@ -1 +1,2 @@
 from .redis_client import redis_client
+from .redis_client import RedisClient
```
stroncod, line 2: The client is a singleton so calling the already instanced redis_client achieve the same.

stroncod (changes requested):
The idea looks good is just the placement I'm hesitant about. I'll do a PR but it will only modified the redis_client, so lets merge that one first (should not clash with this) and move with this one after the changes.

scheduler/services/visibility/calculator.py
dngomez: Sure, I agree, I just started doing the change in the `fill_redis` script to keep the previous logics, actually is pretty much a refactor of the script code.

scheduler/services/redis_client/__init__.py
dngomez: VSCode was complaining, it didn't find RedisClient, maybe just an editor mistake :cry:

stroncod: If is called like `redis_client.flatten_dict` instead of `RedisClient.flatten_dict` ? if not lets add it for peace of mind.

PR #535, opened 2025-01-07
Imaging filters added

dngomez:
New filters.txt files to add imaging filters.
New pickle files.


scheduler/services/visibility/calculator.py
```diff
@@ -153,13 +152,12 @@ class VisibilityCalculator(metaclass=Singleton):
...
 
-    def __init__(self, with_redis: bool = True):
+    def __init__(self):
         self.vis_table: Optional[TargetVisibilityTable] = None
-        self.with_redis = with_redis
 
-    async def calculate(self) -> None:
-        if self.with_redis:
-            all_semesters_vis_table = {}
+    async def calculate(self, with_redis: bool = True, programs_ids: Optional[bytes] = None) -> None:
```
stroncod, line 158: Just curious, why change it from a class variable to a parameter, is there a case when we have the calculator doing both with and without redis in run time?

scheduler/services/visibility/calculator.py
```diff
@@ -169,8 +167,17 @@ async def calculate(self) -> None:
                     _logger.info(f'Visibility calcs for {semester} from Redis retrieved.')
 
         else:
-            # fill the table manually. see fill_redis code.
-            raise NotImplementedError('Manual population not yet available')
+            from scheduler.scripts.fill_redis import compute_visibility, transform_to_json
```
stroncod, line 170: I think is the same code that the other PR but we should not be importing from fill_redis.py as is a standalone script. In this case compute_visibility should be in the calculator and be imported in fill_redis.py , not the other way around. 

I'm guessing this change is to reutilize code? Part of the problem is we have the programprovider and collector mixed too much so that part of the code lives separately. If we use the same compute_visibility method describe here we would be doing the parsing again  when we run the scheduler normally.

stroncod (changes requested):

scheduler/services/visibility/calculator.py
dngomez: Sorry, this is trash code, I was testing locally and I needed a way to avoid retrieving data from Redis. This shouldn't be part of the PR, I will fix it right now.
Same thing with the previous comment.

scheduler/pickles/simresource.pickle
stroncod, line 1: I don't think we have to worry much about the simresource by now.

stroncod (approved):
Looks good, Now the we added gmos filter as a resource a change in the UI is needed I guess. I'm not sure we do have those working tho

scheduler/pickles/simresource.pickle
dngomez: I agree, but implemented the same logic just in case.

dngomez:
I made the filter change in the UI a long time ago, when we found the instrument definition in `lucupy` and we noticed there were some missing instruments.
The current table under the plan plot shows the selected filter, I have checked locally and looks like it was working fine with this PR.

PR #542, opened 2025-03-06
Enabled with_redis boolean

dngomez:
In the `config.yaml` file the parameter `with_redis` will be found and can be used to disable redis.
The visibility calculations are done in the collector and just for the programs needed in the requested plan, which is different from the information retrieved from Redis that stores a full list of programsfor the semester, calculatiing the full list locally at startup means a huge overhead.


bryanmiller (approved):
Your approach sounds good.  Let's try it.

PR #546, opened 2025-04-02
GSCHED-793: Remove TargetInfo unused fields in Collector and Move ephemerides files to a compressed file

stroncod:
This includes changes for the GSCHED-694 task but is not the final solution, is just temporary so we see how it behaves the scheduler without having to download all the ephemerides files. 
This would require **git lfs** to push the ephemerides file as is too big for normal git to handle. Keep that in mind or consecutives pushs would failed after merging.


dngomez (approved):
LGTM!

PR #547, opened 2025-04-03
Small changes for GPP generalization

bryanmiller:
These changes are mostly related to generalizing for future GPP integration
- ranker/default.py (band from obs rather than program)
- ocsprogramprovider (cleanup)
- statscalculator.py (band from obs rather than program)
- file_based_resource_service.py (ignore commented lines in file)
- run.py (ignore commented lines in file, add example for plan output to file)
- nightchanges.py (added option to NightlyTimeline.display() to output to an optional text file for simple checks on output)

### DIFF
--- a/scheduler/services/resource/file_based_resource_service.py
+++ b/scheduler/services/resource/file_based_resource_service.py
@@ -583,35 +583,36 @@ class FileBasedResourceService(ResourceService):
             with open(path, 'r') as input_file:
                 too_activations = self._too_activations[site]
                 for line in input_file:
-                    too_id, too_date, too_time = line.split()
-                    too_datetime = datetime.strptime(too_date+' '+too_time, '%Y-%m-%d %H:%M:%S.%f')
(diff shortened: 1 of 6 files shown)


scheduler/services/resource/file_based_resource_service.py
stroncod, line 1: This change won't be reflected if the resource.pickle file is not eliminated and recreated (by running the scheduler again).

stroncod (approved):
Changes look good my only concern is the resource change might not be working if the pickle file is not build again.

scheduler/services/resource/file_based_resource_service.py
dngomez: I agree with Sergio.
Right now this script is being run only if the `resource.pickle` is not found and a new pickle file will be generated.
In other workds, every time you want to modify the resources txt file `resource.pickle` or the way the txt file is parsed, the `resouce.pickle` should be removed.

dngomez (approved):
LGTM

PR #550, opened 2025-04-09
Add hours_up.py example and timing print statements

bryanmiller:
Added an example of how to calculated the hours a target is below a given airmass between twilights. This includes a routine from vskyutil/thorsky that was not included in lucupy since we weren't using it. 

I also added print statements so that we can time the collector, the selector/ranker, and GreedyMax.

### DIFF
--- a/scheduler/engine/engine.py
+++ b/scheduler/engine/engine.py
@@ -3,6 +3,7 @@
 
 from typing import Dict, Optional, Tuple, Generator
 
+import time
 import numpy as np
 from lucupy.minimodel import Site, NightIndex, VariantSnapshot, TimeslotIndex
 from lucupy.timeutils import time2slots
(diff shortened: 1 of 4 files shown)


stroncod (commented):
Let us discuss about the script before merging

bryanmiller:
Today I removed the timing print statements and made the following changes
- allow setting the Horizons query time slot, if you want it to be the same as the standard time slot. This is kept at 1m, as before.
- Calculate the sidereal proper motion once per night at the middle of the night, to reduce memory usage, and marginally speed up the calculations.

bryanmiller:
This now requires the following lucupy PR: https://github.com/gemini-hlsw/lucupy/pull/119

stroncod:
> This now requires the following lucupy PR: [gemini-hlsw/lucupy#119](https://github.com/gemini-hlsw/lucupy/pull/119)

Change is merged!

stroncod (approved):
Looks good, lets implement these changes so we can reduce memory on the Scheduler

PR #551, opened 2025-04-10
Coords should not be stored after calculations

dngomez:
Checking the entire scripts I notice after the `target_snapshot` calculation, the coords were used only once to get the max and min dec for each night.
I modified to store the single max and min values instead of coords and the memory usage in the last test I ran using full semester visibility, 7 night plan, GN only and full program list, decreased from ~17GB to ~9GB.

### DIFF
--- a/scheduler/core/calculations/targetinfo.py
+++ b/scheduler/core/calculations/targetinfo.py
@@ -46,7 +46,8 @@ class TargetInfo:
     rem_visibility_time is the remaining visibility time for the target for the observation across
     the rest of the time period.
     """
-    coord: SkyCoord
+    max_dec: Angle
+    min_dec: Angle
     alt: Angle
     az: Angle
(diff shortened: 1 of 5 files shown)


stroncod (approved):
Great finding! Looks like it needs a rebase but other than looks good!

bryanmiller (commented):
In terms of memory use this is equivalent to the approach that I took in  https://github.com/gemini-hlsw/scheduler/pull/550. In both we store 2 floats/target/night rather than the the coordinates at each time step. The advantage of the the other method is that it is simpler (one line of code changed) and we may need RA/Dec in future version of the scoring algorithm. So, I have a slight preference for the approach in the other PR.

stroncod:
I will close this for now until Diego is back so we can discuss it later as we merged Bryan change instead

PR #556, opened 2025-04-25
Night config handling

dngomez:
@MarkEverett I was checking your changes and I had the same idea, I was implementing this on the collector side, but it is pretty much the same.
My only concern is when we move to realtime, we would need a way to re-create this `self.night_configurations` filtered data every time something is modified, but for validation this means an improvement.

Lets wait for @stroncod and @bryanmiller feedback.


stroncod (approved):
Seems fine to me.

PR #558, opened 2025-04-29
Coarse time grid bug fixes

bryanmiller:
These fixes should allow us to run the scheduler on an arbitrary integer minute time grid. It's been tested using 5 minute time slots.  To change this in Heroku we need to update time_slot_length in config.yaml and then repopulate the redis cache.


stroncod (approved):
👍

PR #559, opened 2025-05-02
Minor fix to setting observation status during time accounting

bryanmiller:
Set the observation status to ONGOING if is not complete (has been split).


stroncod (approved):
Looks good. I would merge it.

PR #560, opened 2025-05-09
GSCHED-654: Refactor of Event Cycle and New Documentation

stroncod:
First: Sorry for the big PR. The changes are mainly allocated on our Events module. Also the documentations add stuff that makes this seems bigger.

- Event reorganization: the Event Cycle (former loop) and Queue share a package. the Event cycle is modularised and a set of unit test where added for the Cycle.
- The new documentation page adds an initial overview and the Events changes. More change would be added conforming new stuff being added and old being changed.
- Puting some old docstrings to standard, they were a mix of styles and no-styles in there 
- Removing some dependencies in the old structure.


scheduler/core/events/queue/events.py
```diff
@@ -3,6 +3,7 @@
 
 import uuid
 from abc import ABC
+from asyncio import Event
```
dngomez, line 6: It looks like this import is not being used

scheduler/core/events/queue/nightchanges.py
```diff
@@ -85,6 +85,7 @@ def get_final_plan(self, night_idx: NightIndex, site: Site) -> Optional[Plan]:
                         last_visit.time_slots = t - last_start_time_slot
                 t = entry.start_time_slot
             else:
+                #TODO:  Add to the plan what is only accounted for (GSCHED-809)
```
dngomez, line 88: Why this TODO?

scheduler/engine/params.py
bryanmiller, line 1: The comments say that the start/end dates are local times, but the example and code has them in UTC. The schedule UI may interpret these as local, but they have to be translated to UTC before setting the parameters.

dngomez (approved):
LGTM!

scheduler/core/events/queue/nightchanges.py
stroncod: So the PR was not bigger

scheduler/engine/params.py
bryanmiller: I updated the doc string to say that the start/end dates are UT.

PR #569, opened 2025-06-16
Added thread event and fix schedule_id context

dngomez:
Added a thread event.
If connection is closed the event is cleared.
The event is check in the most frequent for loops, then if the event is not set in one of the loops a RuntimeError is raised and the current thread is ended.


dngomez:
Will be replaced in the new architecture

PR #572, opened 2025-07-11
GSCHED-833: ongoing score boost, ranker night_indices bug fix

bryanmiller:
Adds a new factor to increase the score for ongoing observations.

Fixes a bug that improves the ranker performance. The ranker is currently instantiated in engine.build, making ranker.night_indices include all the nights to schedule. Therefore, the ranker was looping over all those nights every time it was called. I added a night_indices parameter to ranker.score_observation so that it would only score the one night being treated by the selector. The ranker instantiation is the same so that the arrays are zeroed and the metric_score is defined for later use. There may be a better way to do this but it was simple and it works. We can do more fundamental refactoring for real-time mode.

### DIFF
--- a/scheduler/core/components/ranker/base.py
+++ b/scheduler/core/components/ranker/base.py
@@ -64,7 +64,8 @@ class Ranker(ABC):
             raise ValueError('Ranker group scoring can only score groups.')
 
     @abstractmethod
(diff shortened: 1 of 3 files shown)


stroncod (approved):
Ranker should do one night at a time by default. Let's hope this would improve performance,

PR #573, opened 2025-07-15
Set NIRI obs to imaging and fixed disperser

dngomez:


stroncod (approved):
There is a print on line 1216 that needs to be removed as well. So if you please can add that. thanks 👍

PR #576, opened 2025-07-17
GSCHED-820: Add gpp-client to GPPProgramProvider

stroncod:


scheduler/core/components/collector/collector.py
dngomez, line 1: Are you sure you want to leave those print statements.
If needed, maybe can be logged using the python logger in debug or info

dngomez (approved):
Approved, just take a look at the print statements.

scheduler/core/components/collector/collector.py
stroncod: No, thanks! I did a find all in Pycharm but it failed me.

PR #588, opened 2025-08-13
Update fill_redis script

dngomez:
Updated fill_redis to use the right Collector parameters.
Probably will change again once the StorageManager is ready.


stroncod (approved):
Looks good now the calculate is on the Collector.

PR #592, opened 2025-09-04
Fix a problem where observation statuses are erroneously labeled ONGOING

MarkEverett:
There are two cases where this is happening: (1) Single observations that begin with a READY obstatus at the start of the visit, but then get interrupted partway through the first atom or acquisition are the ones affected.  These should remain READY since no atom is completed.  (2) Observation groups will have charge_group==False if they are started, but not completed.  However, there doesn't seem to be a reason to mark these ONGOING since they must be observed whole.


stroncod (approved):
LGTM. My only worried might be the propagation of groupvisit to the rest of the output structures (NightlyTimeline and stuff)

PR #593, opened 2025-09-04
Stop time accounting for most events when site is blocked by InterruptionEvent

MarkEverett:
In the ChangeMonitor, WeatherChangeEvents during an interruption should not result in time accounting (time accounting would have been updated by the InterruptionEvent and no observations taken since).  This caused a problem because the end_timeslot_charge was getting set to each new event timeslot and causing observations during the interruption to get counted as observed.  For InterruptionEvents, time accounting should be done if there's not already another InterruptionEvent in place.  Similarly, at morning twilight, the time accounting should be done if the site is not blocked.

In the EventCycle, separate the creation of a final plan from time accounting so that, at morning twilight, when no time accounting should be done (ie. probably because of an InterruptionEvent), a final plan can still be created.

A new problem appears if an InterruptionEvent occurs before EveningTwilightEvent.  An example is Gemini South on UT 2018-10-10.  On that night/site, the InterruptionEvent is 5 seconds before EveningTwilightEvent (both at timeslot 0).  The SortedKeyList where they are stored ensures they are ordered by time.  In that case, and with these changes, no partial plan is created at evening twilight.  In fact, no partial plan is created at all this night and there is not a final plan either (closed all night).  This doesn't seem normal, but I'm unsure of the use of the final plan or what it is intended to represent.  To get accommodate this case, I had to modify the calculate_time_losses in NightlyTimeline to check for no final plan at the MorningTwilightEvent.  On other closed nights that I tested, the closure starts ~1 minute past evening twilight and there is a partial plan made to start the night (and then final plan too, based on that).

### DIFF
--- a/scheduler/core/components/changemonitor/change_monitor.py
+++ b/scheduler/core/components/changemonitor/change_monitor.py
@@ -157,6 +157,7 @@ class ChangeMonitor(SchedulerComponent):
                                     f'timeslot {event_timeslot}.')
                 return TimeCoordinateRecord(event=event,
                                             timeslot_idx=last_timeslot_for_night,
+                                            perform_time_accounting=self.is_site_unblocked(site),
                                             done=True)
 
             case WeatherChangeEvent(variant_change=variant_change):
@@ -164,11 +165,15 @@ class ChangeMonitor(SchedulerComponent):
                 self.selector.update_site_variant(site, variant_change)
 
                 # If the site is blocked, we have no reason to recalculate a plan until all blocking events
-                # are unblocked.
+                # are unblocked.  Also, perform no time accounting.
                 if plans is None:
                     if self.is_site_blocked(site):
                         return None
                     raise ValueError(f'No plans have been created for night {night_idx}.')
+                elif self.is_site_blocked(site):
+                    return TimeCoordinateRecord(event=event,
+                                                perform_time_accounting=False,
+                                                timeslot_idx=event_timeslot)
 
                 # Check if there is a visit running now.
                 plan = plans[site]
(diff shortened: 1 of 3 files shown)


stroncod:
The Evening Twilight is the starting point of the Scheduler so right now if that event does not happen the night gets deleted and it causes issues with the UI and I believe the night gets deleted. This is change that we had to revert because they were complains of the side-effects of removing the EveningTwilight event.

scheduler/core/components/changemonitor/change_monitor.py
```diff
@@ -157,18 +157,23 @@ def process_event(self,
...
                 self.selector.update_site_variant(site, variant_change)
 
                 # If the site is blocked, we have no reason to recalculate a plan until all blocking events
-                # are unblocked.
+                # are unblocked.  Also, perform no time accounting.
                 if plans is None:
                     if self.is_site_blocked(site):
                         return None
                     raise ValueError(f'No plans have been created for night {night_idx}.')
+                elif self.is_site_blocked(site):
```
stroncod, line 173: This is a good addition since we don't force changes when the site is blocked.

scheduler/core/events/cycle/cycle.py
```diff
@@ -240,8 +226,22 @@ def _handle_updates(self,
...
 
-            # Get a new selection and request a new plan if the night is not done
-            if not update.done:
+            # If the night is done, get a final plan and add it to the timeline.
+            # Otherwise, get a new selection and request a new plan
+            if update.done:
+                _logger.debug('Night done. Wrapping up final plan')
+                final_plan = nightly_timeline.get_final_plan(NightIndex(night_idx),
+                                                             site,
+                                                             self.change_monitor.is_site_unblocked(site))
```
stroncod, line 235: I guess this would force to get the final plan even if the site is blocked. Let me test this before I can approve it.

dngomez (approved):
Go ahead and merge.

PR #595, opened 2025-09-09
Fix problem parsing the start/end dates of a few FT programs.

MarkEverett:
This branch is not working in the current scheduler on my laptop!!! It appears like the problem is missing visibility data in the Redis DB.  If the program parsing problem prevented certain programs from being stored there, this might make sense.

Four programs are affected where the note title has no whitespace between the string 'Cycle:' and the months of visibility.  E.g., 'Cycle:Dec-Jan-Feb'.  Also replace undefined value in a few logger statements in the _get_program_dates method.


stroncod (approved):
LGTM. Thanks for the fix!

PR #600, opened 2025-09-12
GSCHED-848: GreedyMax OR groups support 

bryanmiller:
Evaluate group, especially OR group, completion and trim branches of the group tree that are no longer needed within GreedyMax.
Validation mode is modified to handle groups in the same way.
This should be good enough for XT1, for the real multi-night simulation mode we will need to apply a similar evaluation to the post-optimizer time accounting.

Requires: https://github.com/gemini-hlsw/lucupy/pull/124


bryanmiller:
I'm going to merge this so that I can move to the next step. I can remove the unnecessary returns in the next PR.

PR #603, opened 2025-09-16
GSCHED-714: GreedyMax support for cadences

bryanmiller:
This adds GreedyMax support for cadences (non-consecutive AND groups).  There is a new internal active flag on groups that GM uses as part of the evaluation. Only the next group in a cadence is active, the rest are turned off. Once the active step is done, the next steps is activated with a timing window based on the cadence parameters (delays).  The scores and visibility_slot_idx's then have to be modified. I had to make the containing classes mutable in order to do this.

These changes should be sufficient for XT1. For the general SIM mode over multiple nights similar changes have to be made to the time accounting step.

This requires https://github.com/gemini-hlsw/lucupy/pull/125


stroncod (approved):
👍

PR #610, opened 2025-10-01
GSCHED-870: Robust identification of twilights and calibration folder

bryanmiller:
Use the observation calibration role attribute to identify twilights and the group system attribute to help identify the baseline calibration groups. gpp-client had to be updated to return the system group attribute, this the gpp-client version was updated to 25.09.0. This also requires the following lucupy PR, so the lucupy requirement was also updated.

https://github.com/gemini-hlsw/lucupy/pull/126


dngomez (approved):
LGTM

PR #687, opened 2026-03-30
Monorepo

dngomez:
### Changelog

**feature** (1): Added CI/CD pipeline with automated testing, CalVer versioning, and Towncrier changelog management.


stroncod (changes requested):
I think putting the weather service inside repo adds to much to the projects and makes it too difficult to navigate, also in reality it shouldn't be part of the repo, we risk to keep it there forever if no other service is created in the future. 

gpp-client should handle the weather service in the future as well, removing the need use direct query/subscription updates here directly, so a change in the weather service is going to impact the gpp-client not the Scheduler.

I think is better if we follow the  FastAPI full stack template, it uses our same stack: https://github.com/fastapi/full-stack-fastapi-template

Having only the two folders in the root (backend = scheduler and fronetd=web) and the setup files outside for the whole project, and setup files (like dockerfiles) inside allows to separate what is needed for each project but creating the sensation that is one. Adding a third service (even if is needed for the Scheduler) adds to much overhead to the project.

stroncod (approved):
I added some modifications on GitOps, CI/CD and fix some Docker issues. I think at this point it should be ready to be merged. Some actions would failed (like ruff linting) but nothing blocks anything yet. Further in development we should aim to be up all standards an have only ✅

PR #695, opened 2026-04-14
Change default visibility range in realtime mode

dngomez:


stroncod (approved):
Looks good! We need to see how are we going to test the Scheduler moving forward with the files Resource.

dngomez:
> Looks good! We need to see how are we going to test the Scheduler moving forward with the files Resource.

Yes I agree, I will merge this and let's check how we can move forward.
Also, I am thinking if the default scheduler parameters should be taken from the build_params_store :thinking:. that's for another PR.

PR #707, opened 2026-05-25
feat(backend): Add step count in plan visits, fix GPP program parsing

bryanmiller:
Fix GPP program parsing after the last gpp-client changes
- some camelCase to snake_case changes
- many values are getting returned like this <GmosNorthFilter.G_PRIME: 'G_PRIME'>.  So, the parser only takes the string after the '.'

Propagates the sequence step_start and step_count values that are now provided by the GPP ODB. Ocsprogramprovider is also updated to provide these values. The plan Visit now includes step_count_idx (0 indexed first step number) and step_count (number of steps) for the atoms in a visit.  These should be displayed in Schedule so that it is clear to the observer what sequence steps are to be executed. I think it will be clearest to show the first and last _1-indexed_ step numbers to match what is displayed in Explore and Observe.  NightlyTimeline.display gives an example of this.

## Scope

- [ ] Backend (`backend/`)
- [ ] Frontend (`frontend/`)
- [ ] Docs (`docs/`)
- [ ] CI/Infra (`.github/`)

## Checklist


bryanmiller:
Looking ahead to sending the plans to Observe, Javier prefers to receive the 0-indexed first step index and the number of steps. But, it would be even better to send a list of ODB sequence step ids (these are long strings). We may need to review all of this once we support splitting NIR spec observations, depends on how we handle some of the calibration steps.

dngomez (approved):
LGTM

PR #711, opened 2026-05-26
fix(backend): Fix time accounting of incomplete AND groups

stroncod:
### Changelog

**fix** (GSCHED-963): Fix time accounting record in Collector. making incomplete AND groups showing up in final plan

## Scope

- [x] Backend (`backend/`)
- [ ] Frontend (`frontend/`)
- [ ] Docs (`docs/`)
- [ ] CI/Infra (`.github/`)

## Jira

[GSCHED-963](https://noirlab.atlassian.net/browse/GSCHED-963)

## Checklist

- [x] Changelog fragment added to `changelog.d/`
- [x] Commit messages follow conventional commits (`feat(backend):`, `fix(frontend):`, etc.)
- [x] No breaking changes (or `breaking` fragment added)

### DIFF
--- a/backend/scheduler/core/components/collector/collector.py
+++ b/backend/scheduler/core/components/collector/collector.py
@@ -756,10 +756,10 @@ class Collector(SchedulerComponent):
                             slot_atom_start = slot_atom_end - slot_atom_length
 
                         if slot_atom_end < end_timeslot_charge:
-
-                            atom_record = self.time_accountant.get_record(observation, obs_seq[atom_idx])
                             if charge_group:
                                 # Charge to program or partner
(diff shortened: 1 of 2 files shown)


backend/scheduler/core/components/collector/collector.py
dngomez, line 1: I don't get the difference.
Is the self.time_accountant.get_record mutating something?

dngomez (approved):
Maybe I am wrong, but I think the result will be the same as before if the get_record function is not mutating anything.

backend/scheduler/core/components/collector/collector.py
stroncod: yes. below that it changes the atom record

PR #714, opened 2026-05-30
fix(backend): Root group changed to folder/OR, GM pseudo time-accounting updates

bryanmiller:
Investigated the issue that Diego reported with the scoring of G-2026A-0166-Q in dev
https://nationalastro.slack.com/archives/C08UDCUGH5Z/p1779999133835539

The root groups was being treated as an AND group, and thus getting scored even if never considered by GM. Changed this to a 'folder' (OR group with all children needed). I could then get plans created in RT mode.

Looked into GM time accounting and Kristin's suggestion that programs were not getting stopped.
- bad bug found/fixed when using PERCENTAGE for the time buffer  (time * (1 + percent)) instead of (time * percent)
- GM pseudo time accounting was not removing groups when the time was getting used  _update_score now does this.
    We may need to check that all the necessary calibrations are taken, or do this check up front (in the Ranker?)

## Scope

- [ ] Backend (`backend/`)
- [ ] Frontend (`frontend/`)
- [ ] Docs (`docs/`)
- [ ] CI/Infra (`.github/`)

## Checklist

### DIFF
--- a/backend/scheduler/core/components/ranker/default.py
+++ b/backend/scheduler/core/components/ranker/default.py
@@ -268,15 +268,11 @@ class DefaultRanker(Ranker):
         if target_info is None:
             return scores
 
-        # Restrict to nights this observation is actually visible on. Sight populates
-        # target_info only for visible (obs, night) pairs; iterating night_indices
-        # blindly would KeyError on the rest. `scores` is already pre-zero-padded
-        # for every night via _empty_obs_scores, so non-visible nights remain zero.
-        obs_nights = sorted(set(night_indices) & target_info.keys())
-        if not obs_nights:
-            return scores
-
+        # Remaining time in the observation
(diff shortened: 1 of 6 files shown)


backend/scheduler/core/components/ranker/default.py
stroncod, line 1: This would cause issues with Sight implementation I believe. The ranker initialize zero arrays for non-visible nights so does it fail someplace else if we do this process for only the nights we have target_info?

dngomez: Tested this PR, without the changes to this file, I kept the file from main.py and everything works correctly.

dngomez (changes requested):
All good but the default.py ranker changes as Sergio mentioned it is required for Sight to work correctly.

PR #717, opened 2026-06-03
feat(backend): Add odb obscalc subscription

dngomez:
### Changelog

**feature** (GSCHED-917): Add odb-obscalc subscription using gpp-client to decide if trigger new plan calculation

Add obscalc subcription using gpp-client to decide if triggering a new plan calculation is required

## Scope

- [ ] Backend (`backend/`)
- [ ] Frontend (`frontend/`)
- [ ] Docs (`docs/`)
- [ ] CI/Infra (`.github/`)

## Checklist

### DIFF
--- a/backend/scheduler/clients/scheduler_queue_client.py
+++ b/backend/scheduler/clients/scheduler_queue_client.py
@@ -14,7 +14,7 @@ from pydantic import BaseModel, field_serializer, field_validator
 from scheduler.graphql_mid.types import NewPlansRT
 from scheduler.events import NightEvent, OnDemandScheduleEvent
 from scheduler.services import logger_factory
-from scheduler.core.events.queue import WeatherChangeEvent
+from scheduler.core.events.queue import WeatherChangeEvent, ObservationActivationEvent
 
 __all__ = ["SchedulerQueue", "SchedulerEvent"]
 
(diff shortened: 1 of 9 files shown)


backend/scheduler/clients/scheduler_queue_client.py
```diff
@@ -73,6 +73,13 @@ async def add_schedule_event(
                     time=event.time,
                     site=event.site,
                 )
+            elif isinstance(event, ObservationActivationEvent):
```
stroncod, line 76: Seems like it duplicates the use case for the WeatherChange. And in the future the OnDemandEvent (as they can be site specific). We might want to handle this better and try to come up with one setup instead of all this branching

backend/scheduler/core/events/queue/events.py
```diff
@@ -161,6 +160,21 @@ class ToOActivationEvent(InterruptionEvent):
     """
     too_id: ObservationID
 
+
+@final
+@dataclass(frozen=True)
+class ObservationActivationEvent(InterruptionEvent):
```
stroncod, line 166: We need to check this hierarchy as well. Does it really need to Interrupt? Just one type of ToOs interrupt the Scheduler and it can even bypass the whole process. Also are we differentiate a regular new observation vs ToOs?

stroncod (approved):
Lets test it out with real changes

backend/scheduler/clients/scheduler_queue_client.py
dngomez: Yes, I agree.
A better handling can be created, so far what can kill it are the required variables depending on the case, for example in the OnDemand event the night_start is set, but nobody else use / include it.
I will think about a possible improvement here.

backend/scheduler/core/events/queue/events.py
dngomez: So far there is no way to differentiate them, I need to modify the data gotten in the subscription to do that.
Once that's done, I can separate different event types.

PR #723, opened 2026-06-18
fix(backend): Add IGRINS-2 acquisition overhead

bryanmiller:
For now the acq overheads have to be hard coded. These should really come from the ODB query. With the lucupy update the IGRINS-2 tellurics will be scheduled with the science.

## Scope

- [ ] Backend (`backend/`)
- [ ] Frontend (`frontend/`)
- [ ] Docs (`docs/`)
- [ ] CI/Infra (`.github/`)

## Checklist

### DIFF
--- a/backend/scheduler/core/programprovider/gpp/gppprogramprovider.py
+++ b/backend/scheduler/core/programprovider/gpp/gppprogramprovider.py
@@ -919,11 +919,11 @@ class GppProgramProvider(ProgramProvider):
 
         # Disperser
         disperser = None
-        if instrument in ['IGRINS', 'MAROON-X', 'GRACES']:
-            disperser = instrument
+        # if instrument in ['IGRINS', 'MAROON-X', 'GRACES']:
+        #     disperser = instrument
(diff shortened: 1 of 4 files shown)


backend/scheduler/core/programprovider/gpp/gppprogramprovider.py
```diff
@@ -1095,6 +1095,9 @@ def parse_observation(self,
             elif 'FLAM' in mode:
                 if 'SLIT' in mode:
                     acq_overhead = timedelta(seconds=20 * 60)
+            elif "IGRINS" in mode:
+                if "SLIT" in mode:
+                    acq_overhead = timedelta(seconds=7 * 60)
```
dngomez, line 1100: Is this the amount of time the mechanism takes to set a specific position?

dngomez (approved):
LGTM!

backend/scheduler/core/programprovider/gpp/gppprogramprovider.py
bryanmiller: This is the average time to slew and put the object in the slit.

PR #725, opened 2026-06-18
Gsched 982/improve scheduler queue logic

dngomez:
### Changelog

**improvement** (GSCHED-982): Improve the schedule queue event creation and handling

## Scope

- [x] Backend (`backend/`)
- [ ] Frontend (`frontend/`)
- [ ] Docs (`docs/`)
- [ ] CI/Infra (`.github/`)

## Jira

[GSCHED-982](https://noirlab.atlassian.net/browse/GSCHED-982)

## Checklist


backend/scheduler/core/events/queue/events.py
```diff
@@ -100,6 +102,13 @@ class RoutineEvent(Event, ABC):
     ...
 
 
+@dataclass(frozen=True)
+class NightEvent(RoutineEvent):
```
stroncod, line 106: There is a `NightEvents` in calculations as well that we need to connect with this. This information is used extensively throughout the Scheduler (vis calc, night tracker, etc). So we need to have one definition, storage and type

backend/scheduler/core/events/queue/scheduler_queue_client.py
stroncod, line 1: We should move this to `scheduler.core.events.queue` if we are consolidating events on that. This client folder was created when an external MQ was thought to be used for the Queue but is not the case anymore.

backend/scheduler/core/events/queue/scheduler_queue_client.py
```diff
@@ -3,105 +3,45 @@
...
-            scheduler_event = SchedulerEvent(trigger_event=reason)
-        self._queue.put_nowait(scheduler_event)
-        _logger.info(f"Sent Scheduler event: {scheduler_event.trigger_event} ")
-
-    async def consume_events(self, callback: Callable[[SchedulerEvent], NewPlansRT]):
+        if event is None:
+            _logger.error("Attempted to add a None event to the scheduler queue. Event will be ignored.")
+            return
+        self._queue.put_nowait(event)
+        _logger.info(f"Sent Scheduler event: {event.description} ")
```
stroncod, line 32: The event from odb_handler.py raw from the subscription does not have `description` attribute so this would fail I think.

backend/scheduler/night_monitor/event_handlers/odb_event_handler.py
```diff
@@ -42,16 +42,13 @@ async def _on_created_edit(self, event: SchedulerObservationsUpdatesObscalcUpdat
...
             too = event.value.target_environment.first_science_target(include_deleted=False).opportunity()
             if too is not None:
                 # TODO: For now we do nothing until we implement the logic for different types of ToOs.
                 pass # Check the type of opportunity
 
-            await self.scheduler_queue.add_schedule_event(
-                reason=f'New observation edit: {event.edit_type} for {event.value.id}',
-                event=event
-            )
+            await self.scheduler_queue.add_schedule_event(event)
```
stroncod: So we are sending the raw pydantic response? Doesn't this breaks the type hint as `Event` is a class from the Scheduler and not created by the gpp-client. ODB has it own event class also. The other thing is some events are in fact Site specific even if we are scheduling both sites at the time we need a way to not affect the view on Site if its not necessary (e.g A new observation is added in the South while the North night is ongoing, if is not an OR group, the corresponding event should not affect or even trigger a new plan on the North view).

stroncod (commented):
Some questions regarding the new `event` handling. This simplify a lot which is good.

backend/scheduler/core/events/queue/events.py
dngomez: Yes I agree

backend/scheduler/core/events/queue/scheduler_queue_client.py
dngomez: Ok, sounds good to me. I will do it.

backend/scheduler/night_monitor/event_handlers/odb_event_handler.py
dngomez: You are right about this case, since the only one being executed right now is the _on_updated_edit, I didn't consider correctly this case.
I will fix it to use the correct event as in the edit case where ObservationActivationEvent is added to the queue instead.
I will take a deeper look to the Site trigger, so far we are not using them on the plan calculation and that's something we should take care of. Also the event plan time trigger should be considered to compute the plan and not always the start of the night (twilight or custom start).

stroncod (approved):
All good! Thanks for the fixes!

PR #727, opened 2026-06-19
Gsched 974/add support for incoming obs

stroncod:
### Changelog

**feature** (GSCHED-974): Add Sight support to subscriptions and fix configuration for sight

## Scope

- [x] Backend (`backend/`)
- [ ] Frontend (`frontend/`)
- [x] Docs (`docs/`)
- [ ] CI/Infra (`.github/`)

## Checklist

- [x] Changelog fragment added to `changelog.d/`
- [x] Commit messages follow conventional commits (`feat(backend):`, `fix(frontend):`, etc.)

### DIFF
--- a/backend/scheduler/night_monitor/event_handlers/odb_event_handler.py
+++ b/backend/scheduler/night_monitor/event_handlers/odb_event_handler.py
@@ -1,15 +1,23 @@
 # Copyright (c) 2016-2025 Association of Universities for Research in Astronomy, Inc. (AURA)
 # For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
 
+import time
 from datetime import timedelta, datetime, UTC
 from typing import ClassVar, Dict, Tuple, Callable
 
+from scheduler.clients.gpp import gpp
 from scheduler.core.events.queue import ObservationActivationEvent
 from scheduler.night_monitor.event_sources import ODBEventSource
 from .event_handler import EventHandler, LastPlanMock
+from .obscalc_visibility import calculate_and_store_visibility, site_key_from_instrument
+from gpp_client.generated.enums import ObservationWorkflowState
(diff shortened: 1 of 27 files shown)


backend/scheduler/night_monitor/event_listener.py
```diff
@@ -81,12 +81,9 @@ async def _producer(
 
             else:
                 print(f"Listening to {sub_name}")
-                async for data in subscription_factory():
+                async for data in subscription_factory(client):
```
dngomez: Thi client here shouldn't be necessary, it is actually a `None` for the ODBEventSource and ResourceEventSource, it only applies for the weather case, which is handled in the previous if.
Anyways this file was modified in the last merged PR 725 - https://github.com/gemini-hlsw/scheduler/pull/725/changes#diff-d24ede03e738eb818fa127943c68d15c38c91e0221e406cb2ea7d58db76fd198

I would skip this file, since it will cause unnecessary conflicts

backend/scheduler/night_monitor/event_sources.py
dngomez, line 1: same as in the event_listener, this is part of PR 725

backend/scheduler/night_monitor/event_handlers/odb_event_handler.py
dngomez, line 1: I think this would require to resolve manually some conflicts since it was also modified in the PR 725, but the Sight changes LGTM

backend/pyproject.toml
dngomez, line 1: This is Ok to me, but probably will cause conflicts with the last PR.
Main currently has no upper bound.
gpp-dev = ["gpp-client>=26.6.0.dev1"]

dngomez (changes requested):
Looks good to me. I would just remove the `event_listener` and `event_handler` file changes.
And manually resolve the `odb_event_handler` changes to keep the sight changes only.
👍

dngomez (approved):
👍

PR #730, opened 2026-06-25
feat: Modify engineRT to compute plans using event time as start

dngomez:
### Changelog

**improvement** (GSCHED-984): Make the engineRT compute plans starting on current time for events instead of at night start

Also update UI to have both options on demand

The current generated events have a time when they happen that is not used, because every time a plan is calculated the start timeslot used is the twilight or custom night start if exists. Then if a new plan is required starting "now" and "now" is after the night has started, the computed plan starts in the past, otherwise the custom night start must be manually tweaked to match the current time.
The new generated plan will take the event time and if it happens after the night start, the event time will be used to start computing the new plan.

This change may be confusing since the night start used in the server is not displayed.

## Scope

- [x] Backend (`backend/`)
- [x] Frontend (`frontend/`)
- [x] Docs (`docs/`)
- [ ] CI/Infra (`.github/`)

## Jira

[GSCHED-984](https://noirlab.atlassian.net/browse/GSCHED-984)

## Checklist


stroncod (approved):
LGTM

PR #746, opened 2026-07-27
feat(backend): Separate resource/env files for the REALTIME/OPS mode

bryanmiller:
Made a separate set of mock resource/env files so that the operations/realtime files can track reality and the simulation mode can be used for experiments. The files were copied from the simulation mode.  The GN calendar was updated for 26B and the current GMOS IFU configuration.

## Scope

- [ ] Backend (`backend/`)
- [ ] Frontend (`frontend/`)
- [ ] Docs (`docs/`)
- [ ] CI/Infra (`.github/`)

## Checklist


dngomez (approved):
LGTM!

stroncod (approved):
A change needed but we need more work in the future as we are still using some simulation builder stuff in operations

PR #749, opened 2026-08-05
feat(backend): GNIRS/F2 support, schema change updates, improved check of used time

bryanmiller:
GNIRS program parsing
F2 imaging support via the new gpp-client
Fixes related to other schema changes
Check in the Selector whether a group will cause the time used to run over. If so, set scores to 0.

## Scope

- [ ] Backend (`backend/`)
- [ ] Frontend (`frontend/`)
- [ ] Docs (`docs/`)
- [ ] CI/Infra (`.github/`)

## Checklist

### DIFF
--- a/backend/scheduler/core/programprovider/gpp/gppprogramprovider.py
+++ b/backend/scheduler/core/programprovider/gpp/gppprogramprovider.py
@@ -152,7 +152,7 @@ class GppProgramProvider(ProgramProvider):
 
     # Translate instrument names to use the OCS Resources
     _gpp_inst_to_ocs = {'GMOS_NORTH': 'GMOS-N', 'GMOS_SOUTH': 'GMOS-S', 'FLAMINGOS2': 'Flamingos2',
-                        'IGRINS2': 'IGRINS-2', 'GHOST': 'GHOST', 'GNIRS': 'GNIRS'}
+                        'IGRINS2': 'IGRINS-2', 'GHOST': 'GHOST', 'GNIRS': 'GNIRS',}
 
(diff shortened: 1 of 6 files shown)


backend/scheduler/core/programprovider/gpp/gppprogramprovider.py
```diff
@@ -522,9 +515,12 @@ def parse_timing_window(self, data: dict) -> TimingWindow:
...
 
-        # else:
-        #     continue
-        # ToDo: determine how to handle exclusion windows
+        else:
+            # ToDo: determine how to handle exclusion windows
+            start = parsedt("2025-01-01 00:00:00.000000+00:00")
+            repeat = TimingWindow.NON_REPEATING
+            period = TimingWindow.NO_PERIOD
+            duration = timedelta(seconds=500000000)
```
dngomez, line 523: Is the timewindow used to calculate something else?
It is a really big window, about 15 years. I guess the idea is to avoid reaching a border condition if the program doesn't have a window defined.
Is a timewindow always required? Maybe a different flag can be used to describe this situation when no limit is provided 🤔

dngomez: And I think in 15 years we will forget to update this value 😂 if we still are in the observatory

bryanmiller: This is just to make sure that values get set, exclusion windows are not supported yet.

dngomez (approved):
LGTM

PR #751, opened 2026-08-10
Gsched 1008/improve UI to display stitched plans and closure windows

dngomez:
### Changelog

**improvement** (GSCHED-1008): Update UI and server to display every sequence atom, the stitched plan in the UI, closure windows and the trigger event position througout the night

## Scope

- [x] Backend (`backend/`)
- [x] Frontend (`frontend/`)
- [ ] Docs (`docs/`)
- [ ] CI/Infra (`.github/`)

## Jira

[GSCHED-1008](https://noirlab.atlassian.net/browse/GSCHED-1008)

## Checklist


backend/scheduler/engine/engineRT.py
```diff
@@ -217,6 +232,8 @@ async def _compute_event_plan(self, event: Event):
                 values = ti[plans.night_idx].alt[visit.start_time_slot: end_time_slot]
                 alt_degs = [val.dms[0] + (val.dms[1] / 60) + (val.dms[2] / 3600) for val in values]
                 plans.plans[site].alt_degs.append(alt_degs)
+
+            self.nightly_timeline.add(plans.night_idx, site, start_timeslot[site][0], plans.plans[site])
```
stroncod: this would fail as the NightlyTimeline needs the event as well.

backend/scheduler/core/events/cycle/cycle.py
```diff
@@ -258,6 +244,18 @@ def _handle_updates(self,
...
+                # TODO: Not needded if the system returns only stitched plans
+                _logger.debug('Night done. Adding last empty plan to compute night stats.')
+                # final_plan = self._get_final_plan(site, night_idx, nightly_timeline)
+
+                nightly_timeline.add(
+                    NightIndex(night_idx),
+                    site,
+                    current_timeslot,
+                    update.event,
+                    None
```
stroncod, line 257: There is no final plan then? This would probably affect the stats as they are calculated with the final plan in night changes like unschedule time

stroncod (changes requested):
Some changes that might interfere with the system. 

The other thing is the linter is complaining for some unused variable. Nothing major and it should build just fine

backend/scheduler/core/events/cycle/cycle.py
dngomez: Right, no final plan, since every plan is a "full night plan".
The night stats are properly calculated, because the last calculated plan is the same as the final plan. Using this "None" generated plan for the Morning Twilight event forces to the stitched plan to take the plans generated before the Morning Twilight and join them with "None" which is the previous final plan.

backend/scheduler/engine/engineRT.py
dngomez: This is what I want to test and have not been able to do it yet.
You are right about the missing event, I will modify it right now.

stroncod (approved):

